Issue #41, opened 2022-04-07
Fix issue with uploadMetadata and bundlrStorage driver


DataKnox:
**Environment:**
Node v17.2
Ubuntu 20.04 LTS

**Issue:**
When performing the uploadMetadata operation, the Buffer appears to be passed the KeypairIdentityDriver at some point.

**Sample code:**
`
    
    const { Metaplex, MetaplexFile, keypairIdentity, bundlrStorage, mockStorage } = require("@metaplex-foundation/js-next");
    const { Connection, clusterApiUrl, Keypair } = require("@solana/web3.js");
    const fs = require('fs')
    const path = require('path')
    const connection = new Connection(clusterApiUrl("devnet"));
    
    let keypairFile = fs.readFileSync('/home/knox/jsnext/newkey.json')
    const keypair = Keypair.fromSecretKey(Buffer.from(JSON.parse(keypairFile.toString())));
    
    const metaplex = Metaplex.make(connection)
        .use(keypairIdentity(keypair))
        .use(bundlrStorage({
            address: 'https://devnet.bundlr.network',
            providerUrl: 'https://api.devnet.solana.com',
            timeout: 60000,
        }))
    
    
    const artPath = '/home/knox/jsnext/art/'
    async function batchCreator() {
        //const mx = await metaplex();
        const { uri, metadata } = await metaplex.nfts().uploadMetadata({
            name: "My NFT #",
            image: new MetaplexFile(Buffer.from('/home/knox/jsnext/art/1.png'), '1.png'),
            properties: {
                files: [
                    {
                        type: "image/png",
                        uri: new MetaplexFile(Buffer.from('/home/knox/jsnext/art/1.png'), '1.png'),
                    },
                ]
            }
        });
    
        console.log(metadata.image) // https://arweave.net/123
        console.log(metadata.properties.files[0].uri) // https://arweave.net/456
        console.log(uri) // https://arweave.net/789
    }
    
    
    
    batchCreator();
`

Result: 
bigint: Failed to load bindings, pure JS will be used (try npm run rebuild?)
(node:1327) ExperimentalWarning: stream/web is an experimental feature. This feature could change at any time
(Use `node --trace-warnings ...` to show where the warning was created)
node:internal/errors:464
    ErrorCaptureStackTrace(err);
    ^

[...]

PR #42, opened 2022-04-07
fix bundlr from node (#41)

DataKnox:
Related to issue #41 

https://github.com/metaplex-foundation/js-next/issues/41

Strictly typing the IdentityDriver to well, IdentityDriver, breaks the functionality of using the KeypairIdentityDriver. The KeypairIdentityDriver is ideal because I can batch upload and mint NFTs using uploadMetdata(). This pull request does two things:
1) Allows identityDriver to be of type any so that it can be either IdentityDriver or KeypairIdentityDriver
2) Fixes the initialization of the bundlr Client in the NodeBundlr instantiation by using the keypair.secretKey properties in the KeypairIdentityDriver class object

### DIFF
--- a/src/Metaplex.ts
+++ b/src/Metaplex.ts
@@ -13,7 +13,7 @@ import {
   SendTransactionError,
   TransactionError,
 } from '@solana/web3.js';
-import { IdentityDriver, GuestIdentityDriver, StorageDriver, BundlrStorageDriver } from '@/drivers';
+import { GuestIdentityDriver, StorageDriver, BundlrStorageDriver } from '@/drivers';
 import {
   InputOfOperation,
   Operation,
@@ -39,7 +39,7 @@ export class Metaplex {
   public readonly options: MetaplexOptions;
 
   /** Encapsulates the identity of the users interacting with the SDK. */
-  protected identityDriver: IdentityDriver;
+  protected identityDriver: any;
 
   /** Encapsulates where assets should be uploaded. */
   protected storageDriver: StorageDriver;
@@ -73,7 +73,7 @@ export class Metaplex {
     return this.identityDriver;
   }
 
-  setIdentity(identity: IdentityDriver) {
+  setIdentity(identity: any) {
     this.identityDriver = identity;
 
     return this;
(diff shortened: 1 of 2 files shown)


src/Metaplex.ts
```diff
@@ -39,7 +39,7 @@ export class Metaplex {
   public readonly options: MetaplexOptions;
 
   /** Encapsulates the identity of the users interacting with the SDK. */
-  protected identityDriver: IdentityDriver;
```
lorisleiva, line 42: Oh no we don't want to do that haha. If you could just revert the changes of that file, I'll propose another way to fix the type errors.

src/drivers/storage/BundlrStorageDriver.ts
```diff
@@ -126,8 +126,7 @@ export class BundlrStorageDriver extends StorageDriver {
     const bundlr =
       this.metaplex.identity() instanceof WalletAdapterIdentityDriver
         ? new WebBundlr(address, currency, this.metaplex.identity(), options)
-        : new NodeBundlr(address, currency, this.metaplex.identity(), options);
-
+        : new NodeBundlr(address, currency, this.metaplex.identity().keypair.secretKey, options);
```
lorisleiva, line 129: If you want the type to work here on the KeypairIdentityDriver, you simply need to revert the ternary condition like so:

```ts
const identity = this.metaplex.identity();

identity instanceof KeypairIdentityDriver
        ? new NodeBundlr(address, currency, identity.keypair.secretKey, options)
        : new WebBundlr(address, currency, identity, options);
```

DataKnox: This would work so long as the identityDriver object in Metaplex.js is of type 'any'. Since it was set to IdentityDriver, it could never be a KeypairIdentityDriver, and thus, never have access to the keypair.secretKey property.

lorisleiva: KeypairIdentityDriver extends IdentityDriver

DataKnox: Right - ah! I see now. Ok, I made the suggested change (reverted Metaplex.ts back to original state and then changed the bundlr logic to use the ternary function)

lorisleiva: Thanks!

lorisleiva:
Thanks 🍺  I'll publish a new patch today.

PR #48, opened 2022-04-12
Bring back support for Node.js 14

lorisleiva:
Add fetch and AbortController [ponyfills](https://github.com/sindresorhus/ponyfill) to support Node 14.


package.json
```diff
@@ -83,6 +85,6 @@
     "typescript": "^4.5.4"
   },
   "engines": {
-    "node": ">=16.0"
+    "node": "^14.0 || >=16.0"
```
thlorenz, line 88: Why not just `>=14.0`?
Would this also work in Node.js 12? Because if it did then we should make this `>= 12.0` to support the widest range possible.

thlorenz (commented):
Thanks, this will allow more people to allow this lib without friction.

Issue #49, opened 2022-04-12
[chore]: complete a test to verify that data is updated on chain


thlorenz:
Currently tests only check for return value of `create|update Nft` calls which means that they would not break even if those functions wouldn't communicate at all to the programs.

We need to have all tests check for on-chain data. I will add this check to one of the existing tests . This can serve as an example for anyone wanting to do this for the remaining tests as well.

PR #55, opened 2022-04-12
test: ensure createNft results in valid on-chain NFT

thlorenz:
## Summary

This is an example of how we'd verify that an NFT actually is present on-chain vs.
only verifying that our methods _return_ the correct NFT.

The other tests should be updated siimilarly.

FIXES: #49

## Chores

I'm including a few minor changes I applied for issues I ran into while working on this.

- remove unnecessary `async` inside NftClient
- fixing invalid comma in tsconfig
- test util to exit test runner ASAP

### DIFF
--- a/src/modules/nfts/NftClient.ts
+++ b/src/modules/nfts/NftClient.ts
@@ -20,33 +20,31 @@ import {
 } from './operations';
 
 export class NftClient extends ModuleClient {
-  async findNftByMint(mint: PublicKey): Promise<Nft> {
+  findNftByMint(mint: PublicKey): Promise<Nft> {
     return this.metaplex.execute(new FindNftByMintOperation(mint));
   }
 
-  async findNftsByMintList(mints: PublicKey[]): Promise<(Nft | null)[]> {
+  findNftsByMintList(mints: PublicKey[]): Promise<(Nft | null)[]> {
     return this.metaplex.execute(new FindNftsByMintListOperation(mints));
   }
 
-  async findNftsByOwner(owner: PublicKey): Promise<Nft[]> {
+  findNftsByOwner(owner: PublicKey): Promise<Nft[]> {
     return this.metaplex.execute(new FindNftsByOwnerOperation(owner));
   }
 
-  async findNftsByCreator(creator: PublicKey, position: number = 1): Promise<Nft[]> {
+  findNftsByCreator(creator: PublicKey, position: number = 1): Promise<Nft[]> {
     return this.metaplex.execute(new FindNftsByCreatorOperation({ creator, position }));
   }
 
-  async findNftsByCandyMachine(candyMachine: PublicKey, version?: 1 | 2): Promise<Nft[]> {
+  findNftsByCandyMachine(candyMachine: PublicKey, version?: 1 | 2): Promise<Nft[]> {
     return this.metaplex.execute(new FindNftsByCandyMachineOperation({ candyMachine, version }));
   }
 
-  async uploadMetadata(input: UploadMetadataInput): Promise<UploadMetadataOutput> {
+  uploadMetadata(input: UploadMetadataInput): Promise<UploadMetadataOutput> {
     return this.metaplex.execute(new UploadMetadataOperation(input));
   }
 
-  async planUploadMetadata(
-    input: UploadMetadataInput
-  ): Promise<Plan<undefined, UploadMetadataOutput>> {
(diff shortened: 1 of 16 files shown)


test/modules/nfts/createNft.test.ts
```diff
@@ -3,207 +3,215 @@ import spok, { Specifications } from 'spok';
...
-			{
-				address: spokSamePubkey(mx.identity().publicKey),
-				share: 100,
-				verified: true,
-			},
-		],
-		collection: null,
-		uses: null,
-	} as unknown as Specifications<Nft>);
+  // Given we have a Metaplex instance.
```
thlorenz, line 11: I'm realizing that this entire file got reformatted, but this is expected since `.prettierrc` has ` "tabWidth": 2` this is bound to happen.
I'm just wondering why it was still 4-tab spaced .. maybe we should include `./test` for that `lint:fix` task?

src/modules/nfts/NftClient.ts
```diff
@@ -20,33 +20,31 @@ import {
 } from './operations';
 
 export class NftClient extends ModuleClient {
-  async findNftByMint(mint: PublicKey): Promise<Nft> {
```
lorisleiva, line 23: I'm not opposed to removing `async` when it's unnecessary but I usually like to explicitly mark all asynchronous methods as `async` anyway so it's super obvious to developers that these functions will return a Promise rather than having to analyse the return type (if there's even one in the first place).

What do you think about that reasoning? If you disagree, I'm happy to remove them as long as we stay consistent with that decision.

test/modules/nfts/createNft.test.ts
lorisleiva: Agreed. Could you please add the test folder to the lint script on this PR? 🙂

test/modules/nfts/createNft.test.ts
```diff
@@ -3,207 +3,215 @@ import spok, { Specifications } from 'spok';
...
+    collection: null,
+    uses: null,
+  } as unknown as Specifications<Nft>;
+  spok(t, nft, { $topic: 'nft', ...expectedNft });
+
+  // When we then retrieve that NFT.
+  const retrievedNft = await mx.nfts().findNftByMint(nft.mint);
+
+  // Then it matches what createNft returned.
+  spok(t, retrievedNft, { $topic: 'Retrieved Nft', ...expectedNft });
```
lorisleiva, line 59: That's perfect, thanks! ❤️

test/helpers/utils.ts
```diff
@@ -0,0 +1,14 @@
...
+ * Therefore the process hangs for a considerable time after the tests finish, increasing the feedback loop.
+ *
+ * This fixes this by exiting the process as soon as all tests are finished.
+ */
+export function killStuckProcess() {
+  // Don't do this in CI since we need to ensure we get a non-zero exit code if tests fail
+  if (process.env.CI == null) {
+    test.onFinish(() => process.exit(0));
+  }
+}
```
lorisleiva, line 14: I noticed that when I was [playing with Jest a few weeks ago](https://github.com/metaplex-foundation/js-next/blob/loris/jest/tests/helpers/setup.ts#L15-L17). I ended up having a `teardownMetaplex(mx)` method that I would use at the end of my tests. This would close the `_rpcWebSocket` connection like so:

```ts
(mx.connection as any)._rpcWebSocket.close();
```

I wasn't super happy with this solution because we had to access an internal property and we had to teardown each `metaplex` instance. However, it did "properly" kill stuck processes rather than exiting and having to discard CI.

Is there something better we can do? Maybe a hybrid solution that keeps track of all created `metaplex` instance and close them using `onFinish`? What do you think?

src/modules/nfts/NftClient.ts
thlorenz: It's obvious already since the return type is _Promise_ and arguably people taking the step to jump to the declaration should also see that return type.

I've been digging for pros/cons of using `async` when no `await` is inside the function:
https://stackoverflow.com/a/44114047/97443 agrees with what I said and in this thread the other cited reason to use an `async` without an `await` is when the function could potentially throw a synchronous error.

In the end in the above cases it's just cosmetic as long as `v8` does not end up wrapping each result in a promise and unwrapping on the other end (as I thought it did at some point when it sees an `async`).

Whichever way we go, I agree that it's important to be consistent.

PR #56, opened 2022-04-13
chore: upgrade amman to latest version and adapting code to it

thlorenz:
- deps: upgrade amman
- chore: adapt test setup to new amman


lorisleiva (approved):
Noice! Looking clean 🧼

PR #57, opened 2022-04-14
Improve ESM extensions and remove paths

lorisleiva:
This PR aims to make [improve our ESM compatibility](https://gist.github.com/sindresorhus/a39789f98801d908bbc7ff3ecc99d99c#how-can-i-make-my-typescript-project-output-esm) by using import/export that, when transpiled by typescript, use only full relative file paths. It also simplifies the build process by removing TS paths.

## Changelog
- Improve `package.json:exports`
- Remove TS paths and clean up its dependencies (ttsc, ts-transform plugins, etc.)
- Use explicit `.js` extension for every import. — @thlorenz Could we enforce that with our linter?


src/modules/nfts/models/Nft.ts
```diff
@@ -6,12 +6,12 @@ import {
...
 } from '@metaplex-foundation/mpl-token-metadata';
-import { Model } from '@/shared';
-import { MetadataAccount, MasterEditionAccount } from '@/programs/tokenMetadata';
-import { JsonMetadataLoader } from './JsonMetadataLoader';
-import { removeEmptyChars } from '@/utils';
-import { MasterEditionLoader } from './MasterEditionLoader';
-import { JsonMetadata } from './JsonMetadata';
+import { Model } from '../../../shared/index.js';
+import { MetadataAccount, MasterEditionAccount } from '../../../programs/tokenMetadata/index.js';
+import { JsonMetadataLoader } from './JsonMetadataLoader.js';
```
thlorenz, line 11: Anything that's `index.js` doesn't need to be specified explicitly, the directory suffices.
Thus in this and all similar cases `../../../programs/tokenMetadata` would work fine as well.

thlorenz, line 11: It is super uncommon to mention the file extension `.js` in an import or `require`.
I don't understand why this is necessary now .. something seems off if it is.

thlorenz (commented):
We should discuss those odd `.js` imports before deciding to merge.

thlorenz:
> Use explicit .js extension for every import. — @thlorenz Could we enforce that with our linter?

As indicated this is almost never done anywhere and we should avoid the need for this if possible.

lorisleiva:
Agreed, this is bad. Let's find another way. Thanks!

PR #60, opened 2022-04-18
Operation refactoring

lorisleiva:
Refactor operations as tagged unions and operation handlers as simple functions.


README.md
```diff
@@ -291,43 +291,47 @@ As you can see, some of the properties — such as `metadata` — are loaded on
...
+    isPending: () => boolean;
+    isRunning: () => boolean;
+    isCompleted: () => boolean;
+    isSuccessful: () => boolean;
+    isFailed: () => boolean;
+    isCanceled: () => boolean;
+    run: (options?: TaskOptions) => Promise<T>;
+    loadWith: (preloadedResult: T) => Task<T>;
+    reset: () => Task<T>;
+    onStatusChange: (callback: (status: TaskStatus) => unknown) => Task<T>;
```
thlorenz, line 319: If you're not consuming the return value of the callback (which you're probably not) it should have `void` return type instead of `unknown`

README.md
```diff
@@ -291,43 +291,47 @@ As you can see, some of the properties — such as `metadata` — are loaded on
...
-    public isPending(): boolean;
-    public isLoading(): boolean;
-    public isLoaded(): boolean;
-    public isSuccessful(): boolean;
-    public isFailed(): boolean;
-    public isCanceled(): boolean;
+export type Task<T> = {
+    getStatus: () => TaskStatus;
+    getResult: () => T | undefined;
+    getError: () => unknown;
```
thlorenz, line 309: Should this return `Error|string` possibly instead of something totally `unknown`?

README.md
```diff
@@ -291,43 +291,47 @@ As you can see, some of the properties — such as `metadata` — are loaded on
...
+    isRunning: () => boolean;
+    isCompleted: () => boolean;
+    isSuccessful: () => boolean;
+    isFailed: () => boolean;
+    isCanceled: () => boolean;
+    run: (options?: TaskOptions) => Promise<T>;
+    loadWith: (preloadedResult: T) => Task<T>;
+    reset: () => Task<T>;
+    onStatusChange: (callback: (status: TaskStatus) => unknown) => Task<T>;
+    onStatusChangeTo: (status: TaskStatus, callback: () => unknown) => Task<T>;
```
thlorenz, line 320: ditto above

README.md
```diff
@@ -291,43 +291,47 @@ As you can see, some of the properties — such as `metadata` — are loaded on
...
+    onSuccess: (callback: () => unknown) => Task<T>;
```
thlorenz, line 321: ditto above

README.md
```diff
@@ -291,43 +291,47 @@ As you can see, some of the properties — such as `metadata` — are loaded on
...
+    onFailure: (callback: () => unknown) => Task<T>;
```
thlorenz, line 322: ditto above

README.md
```diff
@@ -291,43 +291,47 @@ As you can see, some of the properties — such as `metadata` — are loaded on
...
+    onCancel: (callback: () => unknown) => Task<T>;
```
thlorenz, line 323: ditto above

test/shared/useTask.test.ts
```diff
@@ -0,0 +1,163 @@
...
+    await task.run();
+  } catch (error) {
+    // Then the task is marked as failed and we kept track of the error.
+    t.equal(task.getStatus(), 'failed');
+    t.equal(task.getResult(), undefined);
+    t.equal(task.getError(), error);
+    return;
+  }
+
+  t.fail('Test task should have failed');
```
thlorenz, line 61: I'd remove the `return` from the `catch` block and move this line into the `try` block (as the last line) as it should fail before it reaches it.

test/shared/useTask.test.ts
```diff
@@ -0,0 +1,163 @@
...
+
+  // And an abort controller used to cancel the task.
+  const abortController = new AbortController();
+
+  // When we run the task and abort after 10ms.
+  setTimeout(() => abortController.abort(), 10);
+  try {
+    await task.run({ signal: abortController.signal });
+  } catch (error) {
+    // Fail silently...
```
thlorenz, line 79: Should we assert on the `error` here?

test/shared/useTask.test.ts
```diff
@@ -0,0 +1,163 @@
...
+  }
+
+  // Then the task was marked as canceled.
+  t.equal(task.getStatus(), 'canceled');
+  t.equal(task.getResult(), undefined);
+  t.equal((task.getError() as Event).type, 'abort');
+});
+
+test('it can be reset', async (t: Test) => {
+  // Given a test task that ran successfully.
```
thlorenz, line 89: I'd convert all these comments to `t.comment('Given a task that runs successfully')` since then you see this as part of the test output and if it fails in the middle you know how far it got.

thlorenz (changes requested):
Overall looks good aside from the small nits.
However given the amount of changes I didn't review it all in detail, so mainly assuming that things work as we discussed :)

README.md
lorisleiva: I'm happy to go with `void` but let me explain to you why I went for `unknown` in the first place.

When using arrow functions for one-liners it immediately returns the evaluated expression. That means if you were to use the following listeners, TypeScript would give you a type error.

```ts
let taskStatus = 'pending';
task.onStatusChange(newStatus => taskStatus = newStatus) // Asignment returns the assigned value.

const statusHistory = ['pending'];
task.onStatusChange(newStatus => statusHistory.push(newStatus)) // Push returns the latest index.
```

Instead, you'd need to wrap your one-liner in curly braces for it to work with `void`.

```ts
task.onStatusChange(newStatus => { taskStatus = newStatus })
task.onStatusChange(newStatus => { statusHistory.push(newStatus) })
```

So all in all, not such a big difference but I did find the DX more enjoyable using `unknown`. That being said, I'm happy to go with `void` if you find that it will create less confusion. 🙂

README.md
lorisleiva: I'd love to make that more strict but unfortunately, with JavaScript, people can throw anything they want. In fact, if you try to type-hint a try/catch error, it will tell you it can only be type-hinted with `unknown` or `any`. Whilst, we'll eventually have control over our own `Error` type, we don't control what end-users will be throwing within the task callback which will bubble up into that `getError` method. Therefore, we don't have control over that error type and, sadly, it has to stay `unknown` to account for all possibilities. We could add another generic type with a sensible default like `Task<T, E = Error | string>` but I thought it was a bit overkill. Maybe we can leave it `unknown` for now and revisit later as we implement better error handling and as we get more feedback?

test/shared/useTask.test.ts
lorisleiva: Agreed!

test/shared/useTask.test.ts
lorisleiva: I'd love to have a chat with you about that but I strongly disagree on that one.

I think comments provide the perfect visual separation in tests in a way that gives them breathing room and makes them more enjoyable to read.

Example: With comments.

CleanShot 2022-04-19 at 08 58 27@2x

Without comments. (Not to mention how bad it would look on multi-line comments).

CleanShot 2022-04-19 at 08 59 09@2x

I'd be potentially willing to make the change if `t.comment` was actually providing value but I honestly find the logging of `tape` quite messy (even using something like `tap-spec`) and it just gets lost in a sea of console lines you don't care about when things are successful (that was also part of the reason I started that jest PR that we discarded).

In any case, this is something that would need to be done on a separate PR but I really think the purpose of these comments is not to be logged somewhere, but to **provide a better experience to the person reading the tests**, not reading the logs.

lorisleiva:
See https://github.com/metaplex-foundation/js-next/issues/45

PR #61, opened 2022-04-19
Bundlr fund: expose fund by bytes vs files

cosimo-rip:
Hey Metaplex friends 👋

Having fun playing around with this SDK - it is really handy!

For the bundlr storage option, especially when used with web clients, I think it would be helpful to expose a method to fund a bundlr account for a predicted/known number of bytes in advance. 

The use case I have in mind is for a project with thousands of images (ex a 10k PFP project). It's unlikely that someone in a web client would be able to load all images into memory up front and then pass them to bundlr to be processed as a complete group, but one might know the total number of bytes of all images in advance.

By funding via bytes first, we can process the images in batches, and avoid asking a user on a wallet adapter identity to fund over and over for each batch.

The workflow would be:
- fundBytes
- upload (batch 1), upload (batch 2) ... upload (batch n)

(The fact that you perform balance checks and offer to skip balance checks is actually great for this use case too.)

Worth noting that the placeholders to withdraw remaining funds would not work with batches if done automatically, so it might be useful to make that optional if/when it is introduced.

Cheers!

### DIFF
--- a/src/drivers/storage/BundlrStorageDriver.ts
+++ b/src/drivers/storage/BundlrStorageDriver.ts
@@ -42,7 +42,8 @@ export class BundlrStorageDriver extends StorageDriver {
   }
 
   public async getPrice(...files: MetaplexFile[]): Promise<SolAmount> {
-    const price = await this.getMultipliedPrice(files);
+    const bytes = this.getBytes(files);
+    const price = await this.getMultipliedPrice(bytes);
 
     return SolAmount.fromLamports(price);
   }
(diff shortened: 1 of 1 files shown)


lorisleiva:
Hi there 👋 

Thanks for that, I don't see any reason not to merge this.

> Worth noting that the placeholders to withdraw remaining funds would not work with batches if done automatically, so it might be useful to make that optional if/when it is introduced.

Yes, that's definitely something I need to take into consideration otherwise the pre-funding logic makes no sense. I'm thinking of using a `withdraw` boolean when uploading that will default to `true` to allow users to turn it off when uploading in batches. The only thing is the `fund`/`withdraw` logic is specific to the Bundlr driver so I might need to move the flag within the class instead of the upload method to keep the signatures driver-agnostic.

PR #62, opened 2022-04-19
Start using custom error with human-friendly messages

lorisleiva:


src/errors/BundlrError.ts
```diff
@@ -0,0 +1,51 @@
...
+      ...input,
+      key: `plugin.bundlr.${input.key}`,
+      title: `Bundlr > ${input.title}`,
+      source: 'plugin',
+      sourceDetails: 'Bundlr',
+    });
+  }
+
+  static failedToInitializeBundlr(cause: Error) {
+    return new this({
```
thlorenz, line 15: In general I'm not a huge fan of `this` business. Afaik most linters will complain as well.
To me this would read much clearer if this was `new BundlrError(...`

However I would not even expose errors as static methods since now you cannot easily `instanceof` check them (you worked around this with a `key` property, however the language already provides a mechanism to detect custom error types).

As an example the above I'd implement as follows:

```ts
class FailedToInitializeBundlrError extends BundlrError {
  constructor(cause: Error) {
    super({
      cause,
      title: 'Failed to Initialize Bundlr',
      problem: 'Bundlr could not be initialized.',
      solution:
        'This could happen for a variety of reasons. ' +
        'Check the underlying error below for more information.',
    });
  }
}
```

Which then allows you to do something similar to:

```ts
if (err instanceof FailedToInitializeBundlrError) {
  console.error(err.cause);
}
```

Now you get a separate type for each error which is much closer to how custom errors in JS are supposed to be implemented (see [mdn custom error section](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error#custom_error_types)).

Also in a case that this is _thrown_ it will show up properly as `FailedToInitializeBundlrError` in the output.

thlorenz (changes requested):
In general that looks good, but I feel pretty strongly about providing proper error types instead of functions that do this a way that is somewhat alien to how this is normally done in the language.
Also solita generated proper error types from the Rust errors, so this would all work well together.

src/errors/BundlrError.ts
lorisleiva: Good point! I'll make the changes

src/errors/BundlrError.ts
```diff
@@ -0,0 +1,57 @@
...
+      sourceDetails: 'Bundlr',
+    });
+  }
+}
+
+export class FailedToInitializeBundlrError extends BundlrError {
+  constructor(cause: Error) {
+    super({
+      cause,
+      key: 'failed_to_initialize_bundlr',
```
thlorenz, line 19: What are the keys useful for now?
Given we can detect the specific error via `instanceof` we might not need this anymore?

src/errors/MetaplexError.ts
```diff
@@ -0,0 +1,66 @@
...
+  }
+
+  getFullSource(): string {
+    const capitalizedSource = this.getCapitalizedSource();
+    const sourceDetails = this.sourceDetails ? ` > ${this.sourceDetails}` : '';
+
+    return capitalizedSource + sourceDetails;
+  }
+
+  toString(withName: boolean = true) {
```
thlorenz, line 52: Not sure about this `toString` thingy since it may not print to the console as one would expect when the error goes uncaught, i.e. have you tried `throw new WhateverError(..)` and left it uncaught?
I'm not sure it'll print this, but actually the `message + stack` or similar.

thlorenz (changes requested):
Almost there, just two more nits to figure out :)

src/errors/BundlrError.ts
lorisleiva: I've left it for localisation. It doesn't cost much to do it now vs it will be a pain to update everything later. What do you think?

src/errors/MetaplexError.ts
lorisleiva: I have tested this in a [TypeScript Play](https://www.typescriptlang.org/play?#code/C4TwDgpgBAshwEMwBsIA8CiAnLB7LAyrgK5YDG0AvFAOQDOAJgNY1QA+tAdvAO74vtaWMGVYcaKYgHMAlpzG0weKVgQBbGgG4AUNtCRY8JKkw58ASU5hiwKNQDe2qFCYQQALih1gWOVIA0TlDAMsCont6+nAFBSrgARqhqET5+gc50uMg2MricKVExGSTkEJ5wiCjo2HiEJRTpXvUQACJGMsh0APwFaUFkCMR0ED1QNfiBAL66+tAVxtVmWJbWwADqoQAWJMBEpBR2UADyaqEAPPNVprUrNv60mfsQCvTNbYgddDQAfDraZMgEHQ6IZKiZxlgoOhgBBOAwQRCoI5nFgIAgGHlkCAoJx1GVaJdwUtWNQaITFrUtEFUejMdjXB4vKlojoUWiMZwscFQuEmYVWVAaRyuXFEhBkny-AKhXSmtkQnleizqezZY9SuUjFcIXtStLVZzseqKO8EJ9RpEpSraYaoAMhiNPBC-s4yHlIsQyMB8AAKOSrTVgikWKw2ACUSKCGWIkCwftDwAAdKKkmGBc5gJsZHREwzDgADNRakyJiBLHMAEns-psubck3z6eCWZzITCVCgNaTbdQTcz2eTeDFakOXcHCSSfZbicy8tynFHCZnWRyeSnA+NHbHm-XOc3ps+i9Wy6eB86u8T9uGR9rV4gF6LwIQUg7-dbuAIzKkPoAZghOhAaZBNMQQvsAADCSChP+MgAF4QAwuoUD6YZKlIkbOM4Mg-lAPpviepR2JQpKMAIbAcPhm5EaSwiiBGyKYWywCkAulHNIm3oAKpgLGkHDChTYgYxqLMVgrHTpuADaAAMAC6HG4NxvFAhAKFQAA1M2G7sXQyAyMhACMQHOEJUBgQAYsQyDIEhqmoZK0QYZhbqcN4dpQYgenwYhzSHPhYGQWA0FeQhtkCVGdrurY+7tJ0fkSW8sUgl0UD5lA3xQFWbGnklDZQJ4NBUhFIkse5QWeXBoW+ZpMUfOewG6BmH5fj6PBbAAcninjxLgWRogu1A+MQgFOa6UXuQ6DAAELYoN053lAKX5gAOpwq2QZNUAzZ4WXzYMwx5QVRXCfApU+hFzitR1eKLalkm7QOuJFpMsmpfltA0BG6kXalD2tjyEB5d9jHOCtnDfBltk7fY-nwJZ1lhWGQM-WDENQAACkOSTQ-hKbisjIOpataNEHOiqZTDCVk5wBMg6jGUANJuDj04MrTjF3tN2LAyDNCrTQEXGVA0wgQCQIglNxBwsgWCItCsLwqCCzXPgTkuR6Xq+l2gbKxCtzrFsOy2fREV0DGZY+gxjGJjbXaNIxDKePmkiyJwibxFLDAy4mVZjmz+b25hPb4vmkvS5CGW+0uwcNoHxRPAVLtyDQcdNDldV0AVYde1gKcRZMQumd4CAhGQnaucQP4-vpMiwsAllwnQPrWIk+lM4ylrRPcd6jBCJsnaJC7cDwWlN1bzn7RAqeO7Qchm1XNd1wA+j+ntfKnwcFZY8-V2QtecLYDfwnnhN4xKoeezLkVWQwOK4LYq9wlAQx+M20BxAAbjIDAIVAPD-qgWw917Atz0mQduL1EwBx+rOVc+RaDo1QCpKA6Jb5qHwNAAgRwAAyIJvRQBACUP+AD4AoKfj4bEz4zRuxPphAuAoi6IFLnfYAABNeA5g1BVCLAfBCPoe5OiWP3TCJUxI4ggCPN8lsfp3mnszLg98l4gHgEvGQXCkh1wQrQxim9aDtXvlAdhthOHcM0QwbRmEz4FQACotigD+NEoloDZhYQQ0hajTG8IYImCxxRqYFUQWia8ZBNgQDIEwKA8QEDhKgICGEWAfGB3oQ1ECFCnKZjwCPbOMsISJjnpXXe+965rx9K8IsS9QH6SXgyFOW1L6yyWImTg98jEmI0V4lCQFJjuWACE3CI0GLYX6eXYunAKC4BwuSFWWBhFjVcn1RMyBcDfggDORA4TC7aGmEAA). 🙂

src/errors/BundlrError.ts
thlorenz: Seems superfluous to me, but also doesn't hurt anything I suppose.

PR #64, opened 2022-04-20
Introduce RpcDriver to encapsulate RPC logic and its error handling

lorisleiva:
This PR aims to move RPC logic from the `Metaplex` class into a dedicated `RpcDriver` with only one implementation.

This allows the RpcDriver to be more flexible and more complex (e.g. adding better error handling) without cluttering the Metaplex class.

Key changes here include:
- Moving methods from `Metaplex` to `Web3RpcDriver`
- Expose more data from `confirmTransaction`
- Create special RPC error for failed transactions (for both sending and confirming)
- Keep track of the RPC response for transactions that fail to confirm (this should remove the need for your hack @thlorenz)
- Remove the `sendTransaction` method from the identity driver as it conflicted with the RPC driver and it made more sense to have it lived in the latter.


thlorenz (approved):
Looks good to me overall.
Separating this is a good idea.
I also double checked that now `send/confirm` includes what tools will need for diagnostics purposes.

However the change set is too large for me to check for code correctness, so I'm just gonna trust you on that :)

What I'm confused about is where we now set up the connection config. Is it still via `Metaplex.make(connection)`? 
That would seem odd since we'd be configuring the `rpc` plugin at the top level.

Maybe in that case it should be:

```
Metaplex.make().use(rpc(connection))
```
or something similar?

Not a huge deal, just something that stuck out to me.

PR #71, opened 2022-04-22
Add sendTransaction to WalletAdapterIdentityDriver

cosimo-rip:
I've been doing more work with Bundlr in web clients, and noticed that funding fails with the `WalletAdapterIdentityDriver`, because Bundlr relies on the existence of the `sendTransaction` method:

Screen Shot 2022-04-22 at 2 01 59 PM

The call is here:
https://github.com/Bundlr-Network/js-client/blob/bc2104cf3cc89a5f428828764b818619c4c9a4c1/src/web/currencies/solana.ts#L102

It seemed harmless enough to add this as an additional optional wallet adapter method.

### DIFF
--- a/src/drivers/identity/WalletAdapterIdentityDriver.ts
+++ b/src/drivers/identity/WalletAdapterIdentityDriver.ts
@@ -1,4 +1,4 @@
-import { PublicKey, Transaction } from '@solana/web3.js';
+import { Connection, PublicKey, Transaction, TransactionSignature } from '@solana/web3.js';
 import {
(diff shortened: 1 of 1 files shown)


lorisleiva:
Ah, good shout! I refactored the `sendTransaction` method to be part of a new `RpcDriver` so we have a single point of failure but forgot that this would fail when funding Bundlr. I'll merge that and make one little tweak so we still go through the RpcDriver when sending the transaction. Thank you for reporting this so promptly! 🍺

PR #79, opened 2022-04-26
Move getProgramAccount to RpcDrivers

lorisleiva:
The aim of this PR is to move any usage of `getProgramAccount` or `getMultipleAccountInfos` to the RpcDriver to provide a single point of failure that can be use to catch errors and re-throw them with more context.


thlorenz (approved):
LGTM

PR #83, opened 2022-04-28
feat: init candy machine

thlorenz:
## Summary

Adds the ability to initialize a candy machine from a config as in the below example:

```ts
const mx = await metaplex();
const payer = mx.identity();

const solTreasuryAccount = Keypair.generate();
await amman.airdrop(mx.connection, solTreasuryAccount.publicKey, 100);

const minimalConfig: CandyMachineConfigWithoutStorage = {
  price: 1.0,
  number: 10,
  sellerFeeBasisPoints: 0,
  solTreasuryAccount: solTreasuryAccount.publicKey.toBase58(),
  goLiveDate: '25 Dec 2021 00:00:00 GMT',
  retainAuthority: true,
  isMutable: false,
};

const opts = {
  candyMachine: Keypair.generate(),
};

const cm = mx.candyMachine();
const { transactionId, confirmResponse, candyMachine, ...rest } =
  await cm.initCandyMachineFromConfig(config, opts);
```

Tests have been added for all different configuration options.

## Details

A candy machine module was added with model, operations, operation handlers and transaction
builders and a client, to mirror how this was setup for NFTs.

The config type was tweaked from looking at sugar, our docs and the current candy machine CLI.

## Testing/Diagnostics

To allow better testing the token-metadata and candy machine account providers were exposed to
amman in the `.ammnrc.js` and amman was upgraded.

Useful asserts were added.

I also added the `debug` logging library and use that to log out context to help in diagnosing
issues.

## Remaining Work

Whitelist Settings did not work as expected and we are investigating where the problem lies.
My hunch is that there is a problem with the candy machine program.

Todos were added to make note of that and other remaining work items.


src/modules/candy-machine/transactionBuilders/initCandyMachineBuilder.ts
```diff
@@ -0,0 +1,67 @@
...
+  return TransactionBuilder.make()
+    .add(createCandyMachine)
+    .add({
+      instruction: initCandyMachineIx,
+      signers: [payer],
+      key: 'initializeCandyMachine',
+    });
+}
+
+async function createCandyMachineAccountBuilder(
```
lorisleiva, line 48: I actually don't know how I feel about that whole `create` versus `init` now.

I understand that technically, the Candy Machine is first created by the system program and then initialized by the candy machine program but that should be transparent to our users.

We're exposing too low-level language here.

src/shared/PublicKeyString.ts
```diff
@@ -0,0 +1,52 @@
...
+ */
+export function isValidSolanaAddress(address: string) {
+  return /^[0-9a-zA-Z]{43,88}$/.test(address);
+}
+
+/**
+ * Checks if a string is valid PublicKey address.
+ * @private
+ */
+export function isValidPublicKeyAddress(address: string) {
```
lorisleiva, line 24: I don't like the naming here `isValidSolanaAddress` is a paraphrase of `isValidPublicKeyAddress`.

`isValidSolanaAddress` should be more explicit like `isBase58Address` even though it doesn't really fully test for that either.

IMO this logic is redundant anyway since the `PublicKey` constructor does that for us and the try/catch is more than enough.

src/shared/PublicKeyString.ts
```diff
@@ -0,0 +1,52 @@
...
+ * Tries to convert the {@link value} to a PublicKey.
+ *
+ * @throws {@link AssertionError} if the {@link value} is not a valid base58 string for a PublicKey of a Solana
+ * Account.
+ * @private
+ */
+export function tryConvertToPublickKey(value: PublicKeyString): PublicKey {
+  assert(isPublicKeyString(value), `${value} is not a valid PublicKey`);
+  return new PublicKey(value);
+}
```
lorisleiva, line 52: I thought assertions were supposed to be unexpected program/internal errors. This is very much something that could be exposed to the end-user and should be wrapped in a `MetaplexError` instead.

src/utils/types.ts
```diff
@@ -1 +1,5 @@
-export default {};
+export class UnreachableCaseError extends Error {
+  constructor(value: never) {
+    super(`Unreachable case: ${value}`);
+  }
+}
```
lorisleiva, line 5: This should either be an `AssertionError` or a `MetaplexError` that should be moved to the `errors` folder. I don't want us to support yet another type of error.

test/helpers/amman.ts
```diff
@@ -1,6 +1,55 @@
...
 
 export const amman = Amman.instance({
   knownLabels: { [TOKEN_METADATA_ADDRESS]: 'Token Metadata' },
 });
+
+// -----------------
+// Amman Utils
+// -----------------
+
+// Most of the below should get added to amman or cusper proper
```
lorisleiva, line 16: I want the SDK to support these error resolutions out of the box so we shouldn't even need these helpers in the future. But glad to see them here for now! :)

test/helpers/asserts.ts
```diff
@@ -1,20 +1,71 @@
...
+      // resolution may throw
+      await fnOrPromise();
+      t.fail(`expected to throw ${match.toString()}`);
+    } catch {
+      t.pass(`throws ${match.toString()}`);
+    }
+  } else {
+    return assertThrowsOnResolve(t, fnOrPromise, match);
+  }
+}
```
lorisleiva, line 71: Hmmm, can't we easily combine them using something like this:

```ts
try {
    await (typeof fnOrPromise === 'function' ? fnOrPromise() : fnOrPromise);
} catch (e) {
    // ...
}
```

Also not sure if it's been forgotten but you don't seem to be using the `match` attribute to compare the error.

test/helpers/candy-machine/asserts.ts
```diff
@@ -0,0 +1,23 @@
+import { Creator } from '@metaplex-foundation/mpl-token-metadata';
+import spok from 'spok';
+import { Test } from 'tape';
+import { CreatorsConfig } from '../../../src/modules/candy-machine/models/config';
+
+export function assertCreators(t: Test, creators: Creator[], config: CreatorsConfig) {
```
lorisleiva, line 6: I like the idea of having module-specific assertions on the tests but can we move that to `helpers/candy-machine.ts`. I don't want nested folders in `helpers`.

test/helpers/utils.ts
```diff
@@ -1,4 +1,6 @@
+import { ConfirmOptions } from '@solana/web3.js';
 import test from 'tape';
+import crypto from 'crypto';
```
lorisleiva, line 3: Please no 😅 

Let's avoid node modules at all costs, even in the tests. There are plenty of libraries out there for hashing. We use `tweetnacl` for verifying signatures already, maybe that has something useful for you.

test/modules/candy-machine/initCandyMachine.test.ts
```diff
@@ -0,0 +1,371 @@
...
+import { Signer } from '../../../src/shared';
+import {
+  CandyMachineConfigWithoutStorage,
+  creatorsConfigDefault,
+} from '../../../src/modules/candy-machine/models/config';
+import { assertCreators } from '../../helpers/candy-machine';
+
+killStuckProcess();
+
+async function init() {
```
lorisleiva, line 32: I like this!

test/modules/candy-machine/initCandyMachine.test.ts
```diff
@@ -0,0 +1,371 @@
...
+      retainAuthority,
+      goLiveDate: spokSameBignum(new Date(goLiveDate).valueOf()),
+    },
+  });
+}
+
+// -----------------
+// Minimal Config
+// -----------------
+test('candyMachine: init with minimal config', async (t) => {
```
lorisleiva, line 115: Should we also prefix other tests?
Not sure what's the best convention with tape.

test/modules/candy-machine/initCandyMachine.test.ts
```diff
@@ -0,0 +1,371 @@
...
+  const { cm, solTreasuryAccount, minimalConfig, opts } = await init();
+  const config = minimalConfig;
+
+  const { transactionId, confirmResponse, candyMachine, ...rest } =
```
lorisleiva, line 119: Can I have some Given / When / Then comments please? 😬

test/modules/candy-machine/initCandyMachine.test.ts
```diff
@@ -0,0 +1,371 @@
...
+      mode: 'burnEveryTime',
+      discountPrice: 5,
+      mint: '<invalid mint key>',
+      presale: false,
+    },
+  };
+
+  await assertThrows(
+    t,
+    () => cm.initCandyMachineFromConfig(config, opts),
```
lorisleiva, line 368: There are no tests using `initCandyMachine` directly. I know that `initCandyMachineFromConfig` uses it under the hood but we should at least make sure it works on its own since it's part of the public API.

lorisleiva (changes requested):
Thanks, in general, I like the work that has been done. However, I'm unsure about some of the decisions that were made and I think we might need to have a chat to discuss these.

The main issues I would say would be:
- Lack of consistency with the rest of the SDK.
- API exposed is too low-level.

src/modules/candy-machine/CandyMachineClient.ts
```diff
@@ -0,0 +1,45 @@
...
+      candyMachineModel,
+      confirmOptions,
+    };
+
+    return this.initCandyMachine(input);
+  }
+
+  getCandyMachine(candyMachineAccount: PublicKey) {
+    return CandyMachineModel.getCandyMachine(candyMachineAccount, this.metaplex.connection);
+  }
```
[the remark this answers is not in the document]
thlorenz: I can add another `find` method, but note that his returns a model instead of just the Candy Machine serialized from chain.

src/modules/candy-machine/models/CandyMachine.ts
```diff
@@ -0,0 +1,152 @@
...
+  creatorsConfigDefault,
+  endSettingsFromConfig,
+  gatekeeperFromConfig,
+  hiddenSettingsFromConfig,
+  whiteListMintSettingsFromConfig,
+} from './config';
+import { Connection, PublicKey } from '@solana/web3.js';
+import { getSpaceForCandy } from './candyMachineSpace';
+
+export class CandyMachineModel extends Model {
```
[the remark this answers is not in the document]
thlorenz: I ran into all kind of issues with naming conflicts which is why I changed to model, but I can do that.

src/modules/candy-machine/operationHandlers/initCandyMachineOperationHandler.ts
```diff
@@ -0,0 +1,47 @@
...
+    metaplex: Metaplex
+  ): Promise<InitCandyMachineOutput> {
+    const { payer = metaplex.identity() } = operation.input;
+    const {
+      candyMachine = Keypair.generate(),
+      wallet = payer.publicKey,
+      authority = payer.publicKey,
+      candyMachineModel,
+      confirmOptions,
+    } = operation.input;
```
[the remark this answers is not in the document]
thlorenz: No I cannot since that `payer` is used in the next destruct (so they cannot be in the same one as `payer` would not be assigned yet afaik).
Even if that somehow magically works I'd prefer to be explicit to say:
1. destruct payer
2. use it as default in the next destruct

test/modules/candy-machine/initCandyMachine.test.ts
thlorenz: I like to prefix with the main topic or file name without extension.
It's my convention to quickly find failing tests, but we can do whatever we want.
Tape gives us all the freedom.

test/modules/candy-machine/initCandyMachine.test.ts
thlorenz: Not a huge fan of duplicating tests.
But I can add just one using it directly just to even ensure its exported I suppose.

src/modules/candy-machine/operationHandlers/initCandyMachineOperationHandler.ts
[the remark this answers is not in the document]
lorisleiva: Yes, what I was trying to say is that it does work magically. Destructuring works sequentially. But I don't mind if you want to keep it as-is.

test/modules/candy-machine/initCandyMachine.test.ts
lorisleiva: IMO, `initCandyMachine` should be the one that's heavily tested and `initCandyMachineFromConfig` should have just enough tests to ensure the parsing is correct. That way there's no test duplication, you're testing the specifics of each method.

test/helpers/candy-machine/asserts.ts
thlorenz: These are asserts specific to candy machine, there might be more helpers specific to candy machine.
I'd like to avoid stuffing them all into one large file.

test/helpers/utils.ts
thlorenz: tweetnacl doesn't  have an obvious way to create a 32 bit hash for a message.
Also this is only used / imported in tests which run with Node.js, so I don't thing this is a huge deal.

However if this is something you feel strongly about I'll try to find a browser friendly alternative.

lorisleiva (approved):
lgtm and do some refactoring

Issue #91, opened 2022-05-05
Small candy machine changes


lorisleiva:
As discussed with @thlorenz on a call:

*   Remove `try` prefixes in methods. Not comparable to Rust.
*   Refactor `tryConvertToPublickKey` to something more generic that accepts more value like  `toPublicKey`.
*   Provide a DateTime input type that includes multiple formats like `Date | number | string`.
*   Refactor `UnreachableCaseError` to a `SdkError` which extends `MetaplexError`. Give the unreachable case as parameter.
*   Everything must go through the `rpc()` driver. Using `connection` directly is a smell.
*   Spread the data when creating a candy machine and make as many things as possible optional whilst documenting the defaults value.

PR #94, opened 2022-05-07
chore: small candy machine changes as requested

thlorenz:
## Summary

FIXES: #91 by addressing each requested change

## Details of Addressed Requests

- removed `try` prefixes
- refactor of `convertToPublicKey` put off in order to gain more insight into use cases and
  ideally find an existing implementation instead of hand rolling our own
- provide `DateTime` input for multiple formats also put off for same reasons
- `UnreachableCaseError` is now an `SDKError`
- `findCandyMachine` now uses metaplex rpc
- `CandyMachineData` now is no longer a `data` property but has its fields spread out
  - held off on deciding what's optional + detailed field docs as I want to discover this via
    use cases (that don't provide a candy machine config)

## Additional Changes

- replaced `instanceof` with `is` as that works across web3.js versions for the `PublicKey`
  `equals` method
- properly handling case for a candy machine that was just created, but now cannot be found via
  a custom Metaplex Error created for that case
- exposing candy machine client under `candyMachines`

### DIFF
--- a/src/errors/CandyMachineError.ts
+++ b/src/errors/CandyMachineError.ts
@@ -5,10 +5,10 @@ export class CandyMachineError extends MetaplexError {
   constructor(input: MetaplexErrorInputWithoutSource) {
     super({
       ...input,
-      key: `plugin.candy machine.${input.key}`,
-      title: `NFT > ${input.title}`,
+      key: `plugin.candy_machine.${input.key}`,
+      title: `Candy Machine > ${input.title}`,
       source: 'plugin',
-      sourceDetails: 'NFT',
+      sourceDetails: 'Candy Machine',
     });
   }
 }
--- a/src/errors/CandyMachineError.ts
+++ b/src/errors/CandyMachineError.ts
@@ -1,0 +1,46 @@
+import { PublicKey } from '@solana/web3.js';
+import { MetaplexError, MetaplexErrorInputWithoutSource } from './MetaplexError';
+
+export class CandyMachineError extends MetaplexError {
+  constructor(input: MetaplexErrorInputWithoutSource) {
+    super({
+      ...input,
+      key: `plugin.candy machine.${input.key}`,
+      title: `NFT > ${input.title}`,
+      source: 'plugin',
+      sourceDetails: 'NFT',
+    });
+  }
+}
+
+export class CandyMachineNotFoundError extends CandyMachineError {
+  constructor(candyMachineAddress: PublicKey, cause?: Error) {
+    super({
+      cause,
+      key: 'candy_machine_not_found',
+      title: 'CandyMachine Not Found',
+      problem:
+        'No Metadata account could be found for the provided candy machine address: ' +
+        `[${candyMachineAddress.toBase58()}].`,
+      solution:
+        'Ensure the provided candy machine address is valid and that an associated ' +
+        'Metadata account exists on the blockchain.',
+    });
+  }
+}
+
+export class CreatedCandyMachineNotFoundError extends CandyMachineError {
+  constructor(candyMachineAddress: PublicKey, cause?: Error) {
+    super({
+      cause,
+      key: 'created_candy_machine_not_found',
+      title: 'Created CandyMachine Not Found',
+      problem:
+        'No Metadata account could be found for the candy machine that the client just created: ' +
+        `[${candyMachineAddress.toBase58()}].`,
+      solution:
+        'Ensure that the candy machine could be created properly and without errors.' +
+        'If the problem persists please file an issue.',
+    });
+  }
+}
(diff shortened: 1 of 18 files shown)


src/errors/CandyMachineError.ts
```diff
@@ -0,0 +1,46 @@
...
+import { MetaplexError, MetaplexErrorInputWithoutSource } from './MetaplexError';
+
+export class CandyMachineError extends MetaplexError {
+  constructor(input: MetaplexErrorInputWithoutSource) {
+    super({
+      ...input,
+      key: `plugin.candy machine.${input.key}`,
+      title: `NFT > ${input.title}`,
+      source: 'plugin',
+      sourceDetails: 'NFT',
```
lorisleiva, line 11:
```suggestion
      key: `plugin.candy_machine.${input.key}`,
      title: `Candy Machine > ${input.title}`,
      source: 'plugin',
      sourceDetails: 'Candy Machine',
```

src/plugins/candyMachineModule/createCandyMachine.ts
```diff
@@ -23,11 +23,7 @@ export type CreateCandyMachineOperation = Operation<
   CreateCandyMachineOutput
 >;
 
-export type CreateCandyMachineInput = {
-  // Data.
-  // TODO(loris): spread data into the input directly.
-  candyMachineData: CandyMachineData;
-
+export type CreateCandyMachineInput = CandyMachineData & {
```
lorisleiva, line 26: This is nice but it prevents us from making some fields optional such as the `uuid` that can be inferred by the handler if not provided.

src/plugins/candyMachineModule/CandyMachineClient.ts
```diff
@@ -16,7 +17,7 @@ export type CandyMachineInitFromConfigOpts = {
 };
 
 export class CandyMachineClient extends ModuleClient {
-  findCandyMachineByAddress(address: PublicKey): Promise<CandyMachine> {
+  findCandyMachineByAddress(address: PublicKey): Promise<CandyMachine | null> {
```
lorisleiva, line 20: The NFT module throws "not found" errors on its `findX` methods. Do we think we should return `null` instead? It's a bit problematic when composing multiple handlers together.

src/types/PublicKeyString.ts
```diff
@@ -45,8 +45,9 @@ export function isPublicKeyString(value: string): value is PublicKeyString {
  * @throws {@link AssertionError} if the {@link value} is not a valid base58 string for a PublicKey of a Solana
  * Account.
  * @private
+ * @throws if value is not a valid PublicKey address
  */
-export function tryConvertToPublickKey(value: PublicKeyString): PublicKey {
+export function convertToPublickKey(value: PublicKeyString): PublicKey {
```
lorisleiva, line 50: Maybe not now but it would be nice to include more types here in the future to convert from anything reasonable into a `PublicKey`.

lorisleiva (changes requested):
Made a few comments but thanks for making the changes we discussed.

src/plugins/candyMachineModule/createCandyMachine.ts
thlorenz: Yeah that is planned, just please let me get there once I have a use case for it first.
I hope to run into one when I'm working on the manager.

src/plugins/candyMachineModule/CandyMachineClient.ts
thlorenz: Well the candy machine might not exist, so not sure if we should throw here. Users can check for `null`.

But let's agree on a common pattern and adjust accordingly.

src/types/PublicKeyString.ts
thlorenz: Yeah, as outlined in the PR message I:

> put off in order to gain more insight into use cases and ideally find an existing implementation instead of hand rolling our own

src/plugins/candyMachineModule/createCandyMachine.ts
thlorenz: For instance when managing a Candy Machine all data is required (existing values + updated values), but a missing UUID should _not_ be filled in at that point as the user most likely made an error when not filling it in.

So I'm not ready to flesh out the details here and therefore came up with the simplest solution for now.

src/types/PublicKeyString.ts
lorisleiva: Sorry I missed that

src/plugins/candyMachineModule/CandyMachineClient.ts
lorisleiva: Let's keep it as-is for now and talk about it later to sync the modules together.

lorisleiva (approved):

PR #95, opened 2022-05-08
feat: add candy machine gpa builders

thlorenz:
## Summary

- add GPA builders to find Candy Machines by wallet and authority
- indluded operation handlers exposed via CandyMachine Client


src/programs/candyMachine/gpaBuilders/CandyMachineGpaBuilder.ts
```diff
@@ -0,0 +1,29 @@
...
+  whereDiscriminator(discrimator: AccountDiscriminator) {
+    return this.where(0, Buffer.from(discrimator));
+  }
+
+  candyMachineAccounts() {
+    return this.whereDiscriminator(candyMachineDiscriminator);
+  }
+
+  // wallet same as solTreasury
+  candyMachineAccountsForWallet(wallet: PublicKey) {
```
lorisleiva, line 22: Remind me to have a chat with you about this. I have recently been creating GpaBuilder classes for each account type (that extend a program GpaBuild when needed) where the constructor enforces the account constraint. I think this is good for now but something to discuss when we think about auto-generating them.

lorisleiva (approved):
I love it! Thank you. ❤️

PR #111, opened 2022-05-19
feat: candy machine update authority

thlorenz:
# Summary

Adds `updateAuthority` method to `CandyMachineClient`.

Raising error when updating to same authority since most likely not what the user intended.

FIXES: #108


lorisleiva (approved):
That's great, thanks!

PR #117, opened 2022-05-27
Refactor Storage Module

lorisleiva:
This PR aims to extract some contracts as types instead of classes and bring the SDK's API closer to our goals.

- Create a new `StorageClient` class.
- `StorageDriver` is now only a type required by the `StorageClient`.
- Created and tested `Amount` type with lots of helper methods to manage amounts.
- Removed old `SolAmount`.
- `MetaplexFile` is now only a type and `useMetaplexFile` can be used to create a new file.
- **Removed the plan upload metadata operations** as we're going to support derived keypairs soon.


thlorenz (approved):
I love it. Especially the fact that `Amount` now is only a type as well.
So if I'm not mistaken to implement a storage driver importing sdk _types_ should be sufficient at this point.

PR #119, opened 2022-05-31
Implement Derived Identity Plugin

### LINKED ISSUE #115: Implement derived signer pattern for better web UX
See https://github.com/GlassEaters/signflow/tree/main/packages/derive

lorisleiva:
This PR implements a new plugin to make it easy for developers to use the **derived keypair pattern** with the SDK.

It provides a `derivedIdentity()` client that contains the following methods.

```ts
metaplex.derivedIdentity() // DerivedIdentityClient
metaplex.derivedIdentity().deriveFrom(message) // Promise<void>
metaplex.derivedIdentity().fund(amount) // Promise<void>
metaplex.derivedIdentity().withdraw(amount) // Promise<void>
metaplex.derivedIdentity().withdrawAll() // Promise<void>
metaplex.derivedIdentity().close() // void

metaplex.derivedIdentity().signTransaction(...) // Promise<Transaction>
metaplex.derivedIdentity().signAllTransactions(...) // Promise<Transaction[]>
metaplex.derivedIdentity().signMessage(...) // Promise<Uint8Array>
metaplex.derivedIdentity().verifyMessage(...) // boolean
metaplex.derivedIdentity().equals(...) // boolean
```

As you can see, once the `deriveFrom` method has been called, you can use the `derivedIdentity()` client just like you would use the `identity()` client. You can also fund the derived identity using the current identity and withdraw from it.

Note that, the `derivedIdentity()` client will throw a `UninitializedDerivedIdentityError` when trying to access methods before calling `deriveFrom`.

Fix #115

### DIFF
--- a/test/plugins/derivedIdentity/DerivedIdentityClient.test.ts
+++ b/test/plugins/derivedIdentity/DerivedIdentityClient.test.ts
@@ -135,11 +135,9 @@ test('[derivedIdentity] it can fund the derived identity', async (t: Test) => {
   // When we fund the derived identity by 1 SOL.
   await mx.derivedIdentity().fund(sol(1));
 
-  // And fetch the balances of both the identity and the derived identity.
-  const { identityBalance, derivedBalance } = await getBalances(mx);
-
   // Then we can see that 1 SOL was transferred from the identity to the derived identity.
   // It's a little less due to the transaction fee.
+  const { identityBalance, derivedBalance } = await getBalances(mx);
   t.true(isLessThanAmount(identityBalance, sol(4)));
   t.true(isGreaterThanAmount(identityBalance, sol(3.9)));
   t.true(isEqualToAmount(derivedBalance, sol(1)));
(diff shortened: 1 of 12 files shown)


test/plugins/derivedIdentity/DerivedIdentityClient.test.ts
```diff
@@ -0,0 +1,186 @@
...
+
+test('[derivedIdentity] it derives a Keypair from the current identity', async (t: Test) => {
+  // Given a Metaplex instance using the derived identity plugin.
+  const mx = await init();
+
+  // When we derive the identity using a message.
+  await mx.derivedIdentity().deriveFrom('Hello World');
+
+  // Then we get a Signer Keypair.
+  t.true(mx.derivedIdentity().publicKey);
```
thlorenz, line 60: More common to use `t.ok` and t.notOk`, but those synonyms work fine.
However the failure + success messages will be very non-descriptive.

I'd recommend adding a comment string to the assertion similar to:

```ts
t.true(mx.derivedIdentity().publicKey, 'signer keypair')
```

thlorenz (changes requested):
Would like to see those test assertions updated to make trouble shooting easier.
Aside from that LGTM.

thlorenz (approved):
Thanks. Perfect now :)

PR #123, opened 2022-06-01
Add utils module

lorisleiva:
Adds a utils module that can be used to estimate rent and transaction fees or both.

For now, the transaction fee calculation is trivial but as validators upgrade, we'll need to change that logic to contact the RPC API to figure out the current transaction fee.

This PR also adds an airdrop method to the RPC module.


thlorenz (changes requested):
Looks good, but would like to see some tests if only to see how this is setup + how it is used.

We should be careful not to add too much (non metaplex) functionality since we'll get to maintain it all as well :)

thlorenz (approved):
LGTM

PR #130, opened 2022-06-09
Unspread Candy Machine upload inputs

lorisleiva:
The upload methods were mixing JSON metadata with input data which was making the whole thing difficult to interact with.

This PR wraps the JSON metadata in a nested `metadata` field to fix this.


thlorenz (approved):
Definitely better than before.
I expect this to evolve over time a bit.

PR #131, opened 2022-06-13
Use types and functions instead of classes for handling accounts

lorisleiva:
This PR makes accounts simple types instead of using classes.

For instance:

- A MetadataAccount is now simply defined as: `export type MetadataAccount = Account<Metadata>;`
- `MetadataAccount.pda(...)` now becomes `findMetdataPda(...)` (which can be eventually autogenerated by Shank/Solita)
- `MetadataAccount.from(...)` and `MetadataAccount.fromMaybe(...)` now becomes `parseMetadataAccount(...)` which supports both.
- `parseMetadataAccount(account)` is syntactic sugar for `parseAccount(account, Metadata)`.

It also refactors the very low-level `TransactionBuilder`s into `InstructionWithSigners` so it's closer to how it will look when it will be auto-generated by Solita.


src/programs/candyMachine/accounts/CandyMachineAccount.ts
```diff
@@ -1,62 +1,6 @@
...
+import { CandyMachine } from '@metaplex-foundation/mpl-candy-machine';
+import { Account, getAccountParsingFunction } from '@/types';
 
-export class CandyMachineAccount extends BaseAccount<CandyMachine> {
-  get size() {
-    return getSpaceForCandy(this.data.data);
-  }
+export type CandyMachineAccount = Account<CandyMachine>;
 
-  static getConfigLinesCount(rawData: Buffer) {
```
thlorenz, line 18: I had those as static methods to ease discoverability ... not sure how easy it is now to find those methods given that this package exports tons of things already.
Therefore grouping via class might not be the worst.

Could we find a way to make this easier to discover by grouping exports maybe?

src/programs/tokenMetadata/instructions/createMintNewEditionFromMasterEditionViaVaultProxyInstructionWithSigners.ts
```diff
@@ -0,0 +1,75 @@
+import BN from 'bn.js';
+import { PublicKey } from '@solana/web3.js';
+import { createMintNewEditionFromMasterEditionViaVaultProxyInstruction } from '@metaplex-foundation/mpl-token-metadata';
+import { Signer } from '@/types';
+import { InstructionWithSigners } from '@/utils';
+
+export interface CreateMintNewEditionFromMasterEditionViaVaultProxyInstructionWithSignersParams {
```
thlorenz, line 7: Is this supposed to be extended by package users? Otherwise use `type` here instead of `interface` for consistency.

thlorenz (approved):
Overall LGTM, so I approved.
However as indicated I'm concerned about discoverability of `NFT` vs. `CandyMachine` related methods for example.
They were grouped via classes previously and I'd like to see how we can re-introduce that grouping.

Having everything flat makes it super hard to see what you can do for a specific metaplex _thing_ i.e. NFT vs. Auction vs. CandyMachine.

PR #143, opened 2022-06-21
Update readme funding bundlr 

Steenos:
added info on funding bundlr storage driver


lorisleiva:
That's awesome, thank you for that! ❤️

mPaella:
This is outdated now

Issue #141, opened 2022-06-20
Setting Collection for NFT


myyellowshoe:
Bit confused here, I'm trying to add a collection to a NFT i'm creating, but getting an error. 
`Cannot read property 'toBytes' of undefined`

I have the address using `PublicKey()`, from web3.js but guessing I'm not formatting it properly somehow else? Thanks for any insight here!

myyellowshoe:
Figured out the first part.
Was just a little fuzzy on the needed structure. I imported
`import { Collection } from '@metaplex-foundation/mpl-token-metadata';` and saw the structure from the types.

```
{
verified: false,
key: new PublicKey("mypubkey")
}
```

Next step is to figure out how to verify collection for an nft i've created.

lorisleiva:
Hi there 👋 

I'm closing this since this is now available on the NFT module (See collection parameters on the [create](https://metaplex-foundation.github.io/js/types/js.CreateNftInput.html) and [update](https://metaplex-foundation.github.io/js/types/js.UpdateNftInput.html) operation inputs).

Thank you for raising this.

PR #145, opened 2022-06-23
Verify Collection Feature

jon890:
First, thank you for all metaplex-foundation's convenience library.

Recently, I want to feature that verify collection. and #141 too.
So I refer to Metaplex docs and cli-nft.ts that added verifying collection feature.
There may be something I'm missing.
and I write test code not yet, but simply working well.
Please give us a lot of opinion.

Thanks!

### DIFF
--- a/src/errors/NftError.ts
+++ b/src/errors/NftError.ts
@@ -31,3 +31,19 @@ export class NftNotFoundError extends NftError {
     });
   }
 }
+
+export class CollectionNotFoundError extends NftError {
+  constructor(mint: PublicKey, cause?: Error) {
+    super({
+      cause,
+      key: 'collection_not_found',
+      title: 'Collection Not Found',
+      problem:
+        'No collection could be found for the provided mint address: ' +
+        `[${mint.toBase58()}].`,
+      solution:
+        'Ensure the provided mint address has Collection' +
+        'Collection exists on the blockchain',
+    });
+  }
+}
--- a/src/plugins/nftModule/NftClient.ts
+++ b/src/plugins/nftModule/NftClient.ts
@@ -27,6 +27,11 @@ import {
   PrintNewEditionSharedInput,
   PrintNewEditionViaInput,
 } from './printNewEdition';
+import {
+  VerifyCollectionInput,
+  verifyCollectionOperation,
+  VerifyCollectionOutput,
+} from './verifyCollection';
 
 export class NftClient {
   constructor(protected readonly metaplex: Metaplex) {}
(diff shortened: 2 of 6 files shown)


myyellowshoe:
@jon890 oh snap! Was working on something like this recently for my app, I'm gonna pull this down and check it out.

With this being the `verify Collection` feature this will be assuming the user has already added the collection to the nft metadata and will just need verified.

Once this one get battle tested shouldn't be hard to add the `SetAndVerifyCollectionInstruction` and some of the other functions as well.

lorisleiva:
Yeah, so I want to be careful **not to have a direct one-to-one relationship between the SDK operations and the program instructions**. We should take that opportunity to abstract away some of the on-chain technical necessities.

IMO, we should abstract this within the "updateNft" operation if possible.

Something like this:

```ts
metaplex.nfts().update(nft, {
    collection: PublicKey,
    collectionAuthority: Signer,
})
```

I think the developer experience is much nicer when we create an API that the user would expect. Whilst we can't really do that at the program level, we have that opportunity at the SDK level.

myyellowshoe:
Yeah very great point. Integrating this within the create/update features instead would be much nicer, and easier to use.
Given this though, there is the idea of adding an collection without it being verified. Seams like we'd want some sort of way to signify your adding a collection you want verified, vs one you'd want to verify later.

lorisleiva:
No, I think that's a good idea. Something like that could work:

### Creating NFTs

```ts
{
    collection?: PublicKey,
    collectionAuthority?: Signer
}
```

Examples:

```ts
{} // => No Collection
{ collection: someCollection } // => Unverified Collection
{ collection: someCollection, collectionAuthority: someSigner } // => Verified Collection
```


### Updating NFTs

```ts
{
    collection?: PublicKey | null,
    collectionAuthority?: Signer
}
```

Examples:

```ts
{} // => No changes in Collection
{ collection: null } // => Remove Collection
{ collection: someCollection } // => Set Unverified Collection
{ collection: someCollection, collectionAuthority: someSigner } // => Set Verified Collection
```

lorisleiva:
Hey all 👋  I'm just finishing up a significant refactoring of the NFT module to be more in line with the new modules (that are not yet documented). I should have that merged hopefully today possibly tomorrow. Just a heads up before you invest too much time working with the pre-refactored code.

myyellowshoe:
@lorisleiva sounds great thanks for the heads up!

lorisleiva:
@myyellowshoe Just a quick update: the refactoring has been merged. I just need to spend some time updating the documentation now.

ndrewtl:
Hi @lorisleiva is this feature out with the latest `@metaplex-foundation/js` release? Or is it still in the works?

lorisleiva:
Hey 👋 It's still in the works and we will try to implement it within the `create` and `update` methods instead of adding too many new methods for it to improve the developer experience.

See https://github.com/metaplex-foundation/js/pull/145#issuecomment-1164588253

jon890:
Recently, I started preparing to change company, so I didn't have time to write a better code.
I'm very sorry.
Should I close this pull request?

lorisleiva:
No worries! I think it's best to close this PR anyway since we're planning on implementing this slightly differently. Nevertheless, thank you for your work on the SDK! 🍺

dn-l:
So there's no option in this SDK currently to verify a collection?

PR #182, opened 2022-07-22
Update README.md

makerare:
fix: typo in "create" example code.


lorisleiva:
Thanks! I've got to update that in some other place too so I'll do that as part of another PR.

PR #183, opened 2022-07-22
feat: Added addMetadata to token plugin

stegaBOB:


packages/js/src/plugins/tokenModule/TokenClient.ts
```diff
@@ -45,10 +45,31 @@ import {
...
+  AddMetadataOutput,
+} from './addMetadata';
+import { MintWithMetadata } from '@/plugins';
 
 export class TokenClient {
   constructor(protected readonly metaplex: Metaplex) {}
 
+  addMetadata(
+    input: AddMetadataInput
+  ): Task<AddMetadataOutput & { mintWithMetadata: MintWithMetadata | Mint }> {
```
lorisleiva, line 60: If you're adding metadata and you end up with a `Mint` model then something went wrong.

```suggestion
  ): Task<AddMetadataOutput & { mintWithMetadata: MintWithMetadata }> {
```

You can assert that `mintWithMetadata` is not a `Mint` model by using `assertMintWithMetadata(...)`.

packages/js/src/plugins/tokenModule/TokenClient.ts
```diff
@@ -45,10 +45,31 @@ import {
...
+  AddMetadataInput,
+  addMetadataOperation,
+  AddMetadataOutput,
+} from './addMetadata';
+import { MintWithMetadata } from '@/plugins';
 
 export class TokenClient {
   constructor(protected readonly metaplex: Metaplex) {}
 
+  addMetadata(
```
lorisleiva, line 58: As discussed on Slack, would you mind moving that operation to the NFT module instead?

packages/js/src/plugins/tokenModule/addMetadata.ts
```diff
@@ -0,0 +1,141 @@
...
+  typeof Key,
+  AddMetadataInput,
+  AddMetadataOutput
+>;
+
+export interface AddMetadataInput {
+  // Accounts.
+  mint: PublicKey;
+  payer?: Signer; // Defaults to mx.identity().
+  updateAuthority?: Signer; // Defaults to mx.identity().
```
lorisleiva, line 37: Here the `updateAuthority` does not need to be a Signer.

```suggestion
  updateAuthority?: PublicKey; // Defaults to mx.identity().publicKey.
```

packages/js/src/plugins/tokenModule/addMetadata.ts
```diff
@@ -0,0 +1,141 @@
...
+> & {
+  createMetadataInstructionKey?: string;
+};
+
+export type AddMetadataBuilderContext = Omit<AddMetadataOutput, 'response'>;
+
+export const addMetadataBuilder = async (
+  metaplex: Metaplex,
+  params: AddMetadataBuilderParams
+): Promise<TransactionBuilder<AddMetadataBuilderContext>> => {
```
lorisleiva, line 93: There's no need for this builder method to be async.

lorisleiva (changes requested):
Thanks! I left some comments.

stegaBOB:
Resolved all the comments!

lorisleiva (approved):

lorisleiva:
Thanks!

PR #198, opened 2022-07-27
Add program address and version to Candy Machine model

lorisleiva:
### Problem

Currently, when creating a `CandyMachine` model from the candy machine account, we lose the information regarding "which program" creating this account. This means the consumer cannot know if this Candy Machine used the version 1 or version 2 program.

### Solution

This PR fixes this by adding two more properties to the `CandyMachine` model:
- `programAddress` is the public key of the program that owns the candy machine account.
- `version` can either be `1` or `2` and identifies which Candy Machine program version was used.


lorisleiva:
Published in `v0.13.3`. 🚀

PR #207, opened 2022-08-01
Create Dockerfile

amirsaber:
I am trying to create helm charts for our development cluster. This will greatly help!

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,19 @@
+FROM node:16 as node_builder
+WORKDIR /solana
+RUN wget https://github.com/solana-labs/solana/releases/download/v1.10.32/solana-release-x86_64-unknown-linux-gnu.tar.bz2
+RUN tar jxf solana-release-x86_64-unknown-linux-gnu.tar.bz2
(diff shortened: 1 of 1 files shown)


lorisleiva:
Hi there 👋

Could you please elaborate on how this would be useful?

I'd rather not add the overhead of maintaining Dockerfiles unless there's a significant value to that.

amirsaber:
Hi @lorisleiva 
Vanila Solana test validator doesn't come with metaplex programs loaded. We learned this the hardway and to ease the process for our other devs we are planning to run a loaded test validator in our local cluster which is based on Skaffold and also one in our deployed sandbox k8s cluster. Having the docker image ready to run, hosted on a repo could certainly ease this process. We can certainly fall back into forking and hosting the built image internally if this is just an overhead for you

lorisleiva:
Amman already provides that functionality by wrapping Solana Test Validator and adding more features to it.

I can see the value in having a Dockerfile for that but I don't think it's the responsibility of this repository to maintain it if it's not going to use it internally.

Thank you for your understanding. 🌸

PR #222, opened 2022-08-08
Auction House Cancel Bid/Listing

zaxozhu:
* Implement Cancel Bid operation both for AH & Auctioneer.
* Implement Cancel Listing operation both for AH & Auctioneer.
* Disallow Sale execution if either Sale or Bid is canceled.

### DIFF
--- a/packages/js/src/plugins/auctionHouseModule/AuctionHouseClient.ts
+++ b/packages/js/src/plugins/auctionHouseModule/AuctionHouseClient.ts
@@ -1,6 +1,6 @@
 import { Metaplex } from '@/Metaplex';
 import { now, Signer } from '@/types';
-import { Task } from '@/utils';
+import { Task, Option } from '@/utils';
 import { PublicKey } from '@solana/web3.js';
 import { AuctionHouse } from './AuctionHouse';
 import {
@@ -36,6 +36,16 @@ import {
 } from './findPurchaseByAddress';
 import { LazyPurchase, Purchase } from './Purchase';
 import { LoadPurchaseInput, loadPurchaseOperation } from './loadPurchase';
+import {
+  CancelBidInput,
+  cancelBidOperation,
+  CancelBidOutput,
+} from './cancelBid';
+import {
+  CancelListingInput,
+  cancelListingOperation,
+  CancelListingOutput,
+} from './cancelListing';
 
 type WithoutAH<T> = Omit<T, 'auctionHouse' | 'auctioneerAuthority'>;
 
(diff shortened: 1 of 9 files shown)


packages/js/src/plugins/auctionHouseModule/AuctionHouseClient.ts
```diff
@@ -46,6 +56,56 @@ export class AuctionHouseClient {
...
+
+          return { bid, ...output };
+        } catch (error) {
+          // Fallback to manually creating a listing from inputs and outputs.
+        }
+      }
+
+      return { ...output, bid: null };
+    });
+  }
```
lorisleiva, line 82: Let's not fetch the model after cancelling it here.

I've recently added a design principle on the SDK that **only fetches an account after writing to the cluster if it was created**. Otherwise, we just send the write operation and expect the user to explicitly refresh the model if they need to use the read operations.

So here this should be enough.

```suggestion
  cancelBid(
    input: WithoutAH<CancelBidInput>
  ): Task<CancelBidOutput> {
    return this.metaplex
      .operations()
      .getTask(cancelBidOperation(this.addAH(input)));
  }
```

Same goes for `cancelListing`.

packages/js/src/plugins/auctionHouseModule/cancelBid.ts
```diff
@@ -0,0 +1,152 @@
...
+  if (auctionHouse.hasAuctioneer && !auctioneerAuthority) {
+    throw new AuctioneerAuthorityRequiredError();
+  }
+
+  // Data.
+  const { asset, tradeStateAddress, price, tokens } = bid;
+
+  // Accounts.
+  const tokenAccount =
+    isNftWithToken(asset) || isSftWithToken(asset)
```
lorisleiva, line 88: We've got a boolean for that check.

```suggestion
    !asset.isPublic
```

packages/js/test/plugins/auctionHouseModule/cancelBid.test.ts
```diff
@@ -0,0 +1,171 @@
...
+      mintAccount: nft.address,
+      tokenAccount: nft.token.address,
+      price: sol(1),
+    })
+    .run();
+
+  t.false(bid.canceledAt);
+
+  // When we cancel the given bid.
+  const { bid: canceledBid } = await client.cancelBid({ bid }).run();
```
lorisleiva, line 35: Here we won't return `bid` anymore.

```suggestion
  await client.cancelBid({ bid }).run();
```

This means the test has to then fetch the bid using `findBidByAddress` and make assertions on that.

Same for all other tests.

packages/js/test/plugins/auctionHouseModule/cancelBid.test.ts
```diff
@@ -0,0 +1,171 @@
...
+
+  // Then bid receipt has canceled at date.
+  t.ok(canceledBid?.canceledAt);
+
+  // And the trade state returns the fee to the fee payer.
+  const updatedTradeStateFeeBalance = await mx
+    .rpc()
+    .getBalance((canceledBid as Bid).tradeStateAddress);
+
+  t.equal(updatedTradeStateFeeBalance.basisPoints.toNumber(), 0);
```
lorisleiva, line 45: On other tests, I've done this slightly differently.

```suggestion
  // And the trade state returns the fee to the fee payer.
  const bidAccount = await mx.rpc().getAccount(bid.tradeStateAddress);
  t.false(bidAccount.exists, 'bid account no longer exists');
```

Could you do the same in these tests, please?

packages/js/test/plugins/auctionHouseModule/cancelBid.test.ts
```diff
@@ -0,0 +1,171 @@
...
+
+  // Then the trade state will return the fee to the fee payer.
+  const updatedTradeStateFeeBalance = await mx
+    .rpc()
+    .getBalance(bid.tradeStateAddress);
+
+  t.equal(updatedTradeStateFeeBalance.basisPoints.toNumber(), 0);
+});
+
+test('[auctionHouseModule] cancel a Private Bid on an Auctioneer Auction House', async (t: Test) => {
```
lorisleiva, line 74: Could we also test public bid with auctioneer?

lorisleiva (changes requested):
Thanks that was quick! I've made a few comments as usual. 🙂

lorisleiva (approved):
I've made a few little last comments but all good otherwise. Feel free to merge after tackling these last changes. 🚀

packages/js/test/plugins/auctionHouseModule/cancelBid.test.ts
```diff
@@ -62,7 +62,7 @@ test('[auctionHouseModule] cancel a Public Bid on an Auction House', async (t: T
   // When we cancel the given bid.
   await client.cancelBid({ bid }).run();
 
-  // Then the trade state will return the fee to the fee payer.
+  // And the trade state account no longer exists.
```
lorisleiva, line 65:
```suggestion
  // Then the trade state account no longer exists.
```

packages/js/test/plugins/auctionHouseModule/cancelListing.test.ts
```diff
@@ -71,7 +63,7 @@ test('[auctionHouseModule] cancel a Listing on an Auctioneer Auction House', asy
   // When we cancel the given listing.
   await client.cancelListing({ listing }).run();
 
-  // Then the trade state returns the fee to the fee payer.
+  // And the trade state account no longer exists.
```
lorisleiva, line 66:
```suggestion
  // Then the trade state account no longer exists.
```

lorisleiva (approved):
Very smol changes

PR #228, opened 2022-08-11
Add warning that metaplex does not exist on localhost

mikemaccana:
Heya! New metaplex user, I'm taking a guess - because it doesn't say here - that metaplex doesn't work on localhost with `solana-test-validator`. Is that correct? 

This was me after following the guide, and being used to other programs (like SPL token) that work on localhost:  https://solana.stackexchange.com/questions/1879/metaplex-create-fails-on-localhost-with-attempt-to-load-a-program-that-does-n 

So adding a note to others to make it clear they need to use another network. I hope it's useful!

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,6 +16,9 @@ The entry point to the JavaScript SDK is a `Metaplex` instance that will give yo
 
 It accepts a `Connection` instance from `@solana/web3.js` that will be used to communicate with the cluster.
 
+> **Note**
+> The metaplex program does not exist on localhost. Trying to use it when connected to localhost will fail with `Attempt to load a program that does not exist`. 
+
 ```ts
 import { Metaplex } from "@metaplex-foundation/js";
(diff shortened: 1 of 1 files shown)


lorisleiva:
Hi there 👋 

Sorry for the late reply. So yes only SPL programs exist by default on local test validators.

I don't think we should bother the reader with this information at this stage of the documentation though.

And when we do, we should offer a solution to the reader so they can load Metaplex programs locally. This is easy to do with the native Solana test validator as it allows you to load as many programs as you need via a CLI option.

Internally, we use the [Amman](https://github.com/metaplex-foundation/amman) test library which wraps the Solana test validator and provides additional conveniences such as having a config file at the repository root that loads all of the necessary programs for us.

I'll close this for now, but I'll make sure to include this in the documentation which is being re-written at the moment.

Thank you for raising this!

PR #235, opened 2022-08-15
Add Find Bid by Receipt to Auction House

zaxozhu:
* Adds Find Bid by Receipt.
* Renames findBidByAddress to findBidByTradeState to be consistent and less unclear.


lorisleiva (approved):
Looks good! I'm assuming we're going to do the same for Listings and Purchases.

zaxozhu:
Yea, will do the same :)

PR #250, opened 2022-08-23
Format Amount: fix decimals with leading zeroes

cosimo-rip:
There is a bug that causes formatting of amounts with decimals to lose any leading zeroes.

Ex. 2005000000 lamports becomes "SOL 2.500000000" rather than "SOL 2.005000000"

This is because `mod` is always printed as an integer without leading zeroes. Incorporating decimal length in the toString() call fixes this. Added a test along with the fix to catch this going forward.


lorisleiva:
Hey 👋 Great catch! Thanks for that! 🍺

PR #252, opened 2022-08-23
Add find bids, listings, purchases to Auction House

zaxozhu:
* Adds findBidsBy, findListingsBy, findPurchasesBy to the Auction House module.
* Fixes issue with external seller wallet on createListing.


zaxozhu:
@lorisleiva 

- How do you feel about the interface? 
- Should I add the option to load lazy models?
- It feels that it might be optimized by using getMultipleAccounts somehow.

packages/js/src/plugins/auctionHouseModule/operations/findBidsByPublicKeyField.ts
```diff
@@ -0,0 +1,87 @@
...
+      return await Promise.all(
+        await bidQuery.getAndMap((account) =>
+          metaplex
+            .auctionHouse()
+            .loadBid({
+              lazyBid: toLazyBid(toBidReceiptAccount(account), auctionHouse),
+            })
+            .run(scope)
+        )
+      );
```
lorisleiva, line 85: Yeah I don't think we should load that for them in the default operation handler. This is going to fetch JSON metadata from URI we don't control for many many bids or listings so it's going to be very inefficient.

We should return lazy models instead but allow operation handlers to return normal models.

So the output should be something like this: 

```ts
export type FindBidsByPublicKeyFieldOutput = (Bid | LazyBid)[];
```

That way, if someone wants to write an operation handler that fetches from a web2 database instead of fetching the blockchain, they can return the loaded models directly.

packages/js/src/plugins/auctionHouseModule/operations/findBidsByPublicKeyField.ts
```diff
@@ -0,0 +1,87 @@
...
+  FindBidsByPublicKeyFieldInput,
+  Bid[]
+>;
+
+/**
+ * @group Operations
+ * @category Inputs
+ */
+export type FindBidsByPublicKeyFieldInput = {
+  type: 'buyer' | 'metadata';
```
lorisleiva, line 39: We could easily add `mint` to that list since we can transform get the metadata address from the mint address deterministically.

```suggestion
  type: 'buyer' | 'metadata' | 'mint';
```

lorisleiva (changes requested):
> How do you feel about the interface?

I like it!

> Should I add the option to load lazy models?
> It feels that it might be optimized by using getMultipleAccounts somehow.

I don't think we should load lazy models on these default operation handlers that fetch in bulk from the blockchain. I've commented on that below.

packages/js/src/plugins/auctionHouseModule/operations/findBidsByPublicKeyField.ts
zaxozhu: Thanks, updated!

packages/js/src/plugins/auctionHouseModule/operations/findBidsByPublicKeyField.ts
```diff
@@ -0,0 +1,110 @@
...
+      return await Promise.all(
+        await bidQuery.getAndMap((account) =>
+          metaplex
+            .auctionHouse()
+            .loadBid({
+              lazyBid: toLazyBid(toBidReceiptAccount(account), auctionHouse),
+            })
+            .run(scope)
+        )
+      );
```
lorisleiva, line 108: Let's not offer a `lazy` option since none of the other modules do that and it will be pretty inefficient so I can't see anyone fetching all metadata of all listings/bids at once.

```suggestion
      return bidQuery.getAndMap((account) =>
        toLazyBid(toBidReceiptAccount(account), auctionHouse)
      );
```

lorisleiva (changes requested):

zaxozhu:
@lorisleiva updated!

packages/js/src/plugins/auctionHouseModule/operations/findBidsByPublicKeyField.ts
```diff
@@ -0,0 +1,90 @@
...
+  auctionHouse: AuctionHouse;
+  publicKey: PublicKey;
+  commitment?: Commitment;
+};
+
+/**
+ * @group Operations
+ * @category Outputs
+ */
+export type FindBidsByPublicKeyFieldOutput = LazyBid[];
```
lorisleiva, line 50: Apologies for the confusion but this should still allow for `Bid` to be returned.

```suggestion
export type FindBidsByPublicKeyFieldOutput = (LazyBid | Bid)[];
```

This is because, whilst _our_ operation handlers only return `LazyBid`s other handlers could return `Bid`s directly when fetching from an indexer for example.

packages/js/test/plugins/auctionHouseModule/findBidsByPublicKeyField.test.ts
```diff
@@ -0,0 +1,160 @@
...
+  metaplex,
+  createNft,
+  createWallet,
+} from '../../helpers';
+import { createAuctionHouse } from './helpers';
+import { LazyBid } from '@/plugins';
+
+killStuckProcess();
+
+test('[auctionHouseModule] find all lazy bids by buyer', async (t) => {
```
lorisleiva, line 15: On each of these tests, I like that you create 2 listings/bids/purchases to ensure both are picked up but could you also add a 3rd one that should not be picked up to ensure that we are indeed filtering out the accounts we don't want?

lorisleiva (changes requested):

packages/js/test/plugins/auctionHouseModule/findBidsByPublicKeyField.test.ts
zaxozhu: Thanks, updated!

lorisleiva (changes requested):
One small comment and we should be good. :)

packages/js/src/plugins/auctionHouseModule/operations/createListing.ts
```diff
@@ -269,10 +269,13 @@ export const createListingBuilder = (
...
 
   // Signers.
-  const sellSigners = [seller, authority, params.auctioneerAuthority].filter(
+  const sellSigners = [authority, params.auctioneerAuthority].filter(
     (input): input is Signer => !!input && isSigner(input)
   );
 
+  // Fee payer.
+  const feePayer = isSigner(seller) ? seller : metaplex.identity();
+
```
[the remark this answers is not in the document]
zaxozhu: The problem is that seller is not a signer in this instruction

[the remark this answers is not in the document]
zaxozhu: In create bid it works well, but not here, since it doesn't need seller to be a signer, but only the wallet that runs that transaction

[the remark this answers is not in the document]
zaxozhu: @lorisleiva Maybe it's worth adding a payer input?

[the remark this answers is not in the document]
lorisleiva: Sorry, I'm confused. I thought, in order to list an asset, either:
- The seller has to sign the instruction
- OR The authority of the auction house has to sign it (depending on its settings).

Which is what we had before. I'm not sure why we now always remove the seller as a Signer?

lorisleiva (approved):
Thanks!

PR #258, opened 2022-08-25
Require less data in transaction builders in Auction House

zaxozhu:


packages/js/src/plugins/auctionHouseModule/operations/cancelBid.ts
```diff
@@ -105,28 +118,42 @@ export const cancelBidBuilder = (
...
-
   // Accounts.
   const tokenAccount = isPublic
     ? findAssociatedTokenAccountPda(
         asset.mint.address,
-        toPublicKey(bid.buyerAddress)
+        toPublicKey(buyerAddress)
       )
-    : bid.asset.token.address;
+    : (asset as SftWithToken | NftWithToken).token.address;
```
lorisleiva, line 148: Why do we need the type hint here?

lorisleiva (changes requested):
Just a little comment but otherwise we're looking good. 🙂

packages/js/src/plugins/auctionHouseModule/operations/cancelBid.ts
zaxozhu: TypeScript doesn't see typing that comes from isPublic, so It doesn't treat assets as SftWithToken | NftWithToken by default

lorisleiva: Thanks!

lorisleiva (approved):

PR #265, opened 2022-08-31
Add Deposit instruction

zaxozhu:
* Implement Deposit for Auctioneer AH and simple AH.
* Refactor tests a bit.


packages/js/src/plugins/auctionHouseModule/operations/deposit.ts
```diff
@@ -0,0 +1,188 @@
...
+export const depositBuilder = (
+  metaplex: Metaplex,
+  params: DepositBuilderParams
+): TransactionBuilder<DepositBuilderContext> => {
+  // Data.
+  const auctionHouse = params.auctionHouse;
+  const amountBasisPoint = params.amount.basisPoints;
+  const depositAmount = auctionHouse.isNative
+    ? lamports(amountBasisPoint)
+    : amount(amountBasisPoint, auctionHouse.treasuryMint.currency);
```
lorisleiva, line 116: This is redundant since you're just passing the basis points below on the arguments.

packages/js/src/plugins/auctionHouseModule/operations/deposit.ts
```diff
@@ -0,0 +1,188 @@
...
+ * @group Operations
+ * @category Constructors
+ */
+export const depositOperation = useOperation<DepositOperation>(Key);
+
+/**
+ * @group Operations
+ * @category Types
+ */
+export type DepositOperation = Operation<
```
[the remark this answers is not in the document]
lorisleiva: In fact, we should probably be slightly more explicit on the client methods as well since we'll be withdrawing from other sources too (fee and treasury accounts).

So instead of `metaplex.auctionHouse().deposit(...)`, let's do `metaplex.auctionHouse().depositToBuyerAccount(...)`. And then we'll have `metaplex.auctionHouse().withdrawFromX(...)` where `X` will be `TreasuryAccount`, `FeeAccount` and `BuyerAccount`.

zaxozhu:
@lorisleiva Updated

packages/js/src/plugins/auctionHouseModule/operations/depositToBuyerAccount.ts
```diff
@@ -0,0 +1,202 @@
...
+    | 'authorityAddress'
+    | 'hasAuctioneer'
+    | 'isNative'
+    | 'treasuryMint'
+    | 'feeAccountAddress'
+  >;
+  buyer?: PublicKey | Signer; // Default: identity
+  authority?: PublicKey | Signer; // Default: auctionHouse.authority
+  auctioneerAuthority?: Signer; // Use Auctioneer ix when provided
+  amount: SolAmount | SplTokenAmount;
```
lorisleiva, line 65: Could we start documenting these inputs, outputs and models as we implement them?

We can plan for backfilling all of that documentation later but it would be good if we could at least write documentation as we add new operations at the moment so we have less to do later.

Here's an example of a documented operation:

https://github.com/metaplex-foundation/js/blob/e4c74c22ef4fc7853821f72359f6d82b4bbf0029/packages/js/src/plugins/nftModule/operations/deleteNft.ts

Notice how the [operation constant](https://github.com/metaplex-foundation/js/blob/e4c74c22ef4fc7853821f72359f6d82b4bbf0029/packages/js/src/plugins/nftModule/operations/deleteNft.ts#L16-L28) and the [builder function](https://github.com/metaplex-foundation/js/blob/e4c74c22ef4fc7853821f72359f6d82b4bbf0029/packages/js/src/plugins/nftModule/operations/deleteNft.ts#L120-L132) have a small description and a little example of usage to help users get started.

lorisleiva (changes requested):
All good, just need a little bit of documenting so we have less to catch up on later. 🙂

packages/js/src/plugins/auctionHouseModule/operations/depositToBuyerAccount.ts
zaxozhu: Thanks for the reference. Will do

packages/js/src/plugins/auctionHouseModule/operations/depositToBuyerAccount.ts
```diff
@@ -29,6 +29,15 @@ import { AuctioneerAuthorityRequiredError } from '../errors';
 const Key = 'DepositToBuyerAccountOperation' as const;
 
 /**
+ * Adds funds to user's buyer escrow account for the auction house.
```
lorisleiva, line 32:
```suggestion
 * Adds funds to the user's buyer escrow account for the given auction house.
```

packages/js/src/plugins/auctionHouseModule/operations/depositToBuyerAccount.ts
```diff
@@ -59,9 +69,21 @@ export type DepositToBuyerAccountInput = {
...
+  buyer?: PublicKey | Signer;
+  /**
+   * The Auction House authority.
+   *
+   * @defaultValue `auctionHouse.authority`
+   */
+  authority?: PublicKey | Signer;
+  /** The Auctioneer authority. It is required when Auction House has Auctioneer enabled. */
+  auctioneerAuthority?: Signer;
+  /** Amount of funds to deposit. Can be SOL or any SPL token. */
```
lorisleiva, line 86: A few comments here:
- Let's add an empty line between each parameter when documented so docblocks are visually assigned to their parameter.
- Let's add default values to every single optional parameter.
- Could you explain a bit more why `buyer` and `authority` can be either `PublicKey`s or `Signer`s?

```suggestion
  /**
   * The buyer who deposits funds.
   * TODO: Explain why public key or signer.
   *
   * @defaultValue `metaplex.identity()`
   */
  buyer?: PublicKey | Signer;
  
  /**
   * The Auction House authority.
   * TODO: Explain why public key or signer.
   *
   * @defaultValue `auctionHouse.authority`
   */
  authority?: PublicKey | Signer;
  
  /** 
   * The Auctioneer authority.
   * It is required when Auction House has Auctioneer enabled.
   *
   * @defaultValue Defaults to not being used.
   */
  auctioneerAuthority?: Signer;
  
  /**
   * Amount of funds to deposit. 
   * This can either be in SOL or in the SPL token used by the Auction House as a currency.
   */
```

packages/js/src/plugins/auctionHouseModule/operations/depositToBuyerAccount.ts
```diff
@@ -112,19 +134,28 @@ export type DepositBuilderParams = Omit<
...
  * @category Contexts
  */
-export type DepositBuilderContext = Omit<
+export type DepositToBuyerAccountBuilderContext = Omit<
   DepositToBuyerAccountOutput,
   'response'
 >;
 
 /**
+ * Adds funds to user's buyer escrow account for the auction house.
```
lorisleiva, line 143:
```suggestion
 * Adds funds to the user's buyer escrow account for the given auction house.
```

packages/js/src/plugins/auctionHouseModule/operations/getBuyerBalance.ts
```diff
@@ -32,7 +41,9 @@ export type GetBuyerBalanceOperation = Operation<
  * @category Inputs
  */
 export type GetBuyerBalanceInput = {
+  /** The Auction House in which to get buyer's escrow balance. */
   auctionHouse: Pick<AuctionHouse, 'address'>;
+  /** Buyer's address. */
   buyerAddress: PublicKey;
```
lorisleiva, line 47:
```suggestion
  /** The Auction House in which to get the buyer's escrow balance. */
  auctionHouse: Pick<AuctionHouse, 'address'>;
  
  /** The address of the buyer. */
  buyerAddress: PublicKey;
```

packages/js/src/plugins/auctionHouseModule/operations/getBuyerBalance.ts
```diff
@@ -32,7 +41,9 @@ export type GetBuyerBalanceOperation = Operation<
  * @category Inputs
  */
 export type GetBuyerBalanceInput = {
+  /** The Auction House in which to get buyer's escrow balance. */
   auctionHouse: Pick<AuctionHouse, 'address'>;
```
lorisleiva, line 45: Since we only need the address here let's just require the Public Key directly.

```suggestion
  auctionHouse: PublicKey;
```

lorisleiva (changes requested):

packages/js/src/plugins/auctionHouseModule/operations/depositToBuyerAccount.ts
```diff
@@ -0,0 +1,255 @@
...
+ * @group Transaction Builders
+ * @category Contexts
+ */
+export type DepositToBuyerAccountBuilderContext = Omit<
+  DepositToBuyerAccountOutput,
+  'response'
+>;
+
+/**
+ * Adds funds to user's buyer escrow account for the auction house.
```
lorisleiva, line 164:
```suggestion
 * Adds funds to the user's buyer escrow account for the given auction house.
```

packages/js/src/plugins/auctionHouseModule/operations/depositToBuyerAccount.ts
```diff
@@ -0,0 +1,255 @@
...
+    depositInstruction = createAuctioneerDepositInstruction(
+      { ...accountsWithAuctioneer },
+      args
+    );
+  }
+
+  // Signers.
+  const depositSigners = [buyer, params.auctioneerAuthority].filter(
+    (input): input is Signer => !!input && isSigner(input)
+  );
```
lorisleiva, line 243:
```suggestion
  const depositSigners = [buyer, params.auctioneerAuthority].filter(isSigner);
```

packages/js/src/plugins/auctionHouseModule/operations/depositToBuyerAccount.ts
```diff
@@ -0,0 +1,255 @@
...
+    | 'treasuryMint'
+    | 'feeAccountAddress'
+  >;
+  /**
+   * The buyer who deposits funds.
+   * This expects a Signer.
+   *
+   * @defaultValue `metaplex.identity()`
+   */
+  buyer?: Signer;
```
lorisleiva, line 78: This was `Signer | PublicKey` before. Am I right to assume that the previous implementation was wrong and the buyer always needs to sign?

packages/js/src/plugins/auctionHouseModule/operations/depositToBuyerAccount.ts
```diff
@@ -0,0 +1,255 @@
...
+
+  /**
+   * The Auction House authority.
+   *
+   * @defaultValue `auctionHouse.authority`
+   */
+  authority?: PublicKey;
```
lorisleiva, line 85: If this no longer needs to be a `Signer`, can we not just use `auctionHouse.authorityAddress` directly since it's required in the `Pick` type above?

packages/js/src/plugins/auctionHouseModule/operations/getBuyerBalance.ts
```diff
@@ -0,0 +1,74 @@
...
+
+/**
+ * @group Operations
+ * @category Inputs
+ */
+export type GetBuyerBalanceInput = {
+  /** The Auction House in which to get buyer's escrow balance. */
+  auctionHouse: PublicKey;
+  /** Buyer's address. */
+  buyerAddress: PublicKey;
```
lorisleiva, line 46:
```suggestion
  /** The Auction House in which to get the buyer's escrow balance. */
  auctionHouse: PublicKey;

  /** The buyer's address. */
  buyerAddress: PublicKey;
```

packages/js/src/plugins/auctionHouseModule/operations/getBuyerBalance.ts
```diff
@@ -0,0 +1,74 @@
...
+import { findAuctionHouseBuyerEscrowPda } from '../pdas';
+
+// -----------------
+// Operation
+// -----------------
+
+const Key = 'GetBuyerBalanceOperation' as const;
+
+/**
+ * Gets buyer's balance in Auction House escrow account.
```
lorisleiva, line 13:
```suggestion
 * Gets the balance of a buyer's escrow account for a given Auction House.
```

lorisleiva (changes requested):

packages/js/src/plugins/auctionHouseModule/operations/depositToBuyerAccount.ts
zaxozhu: Yeah, I checked it once more. Buyer must be a signer

lorisleiva (approved):

PR #274, opened 2022-09-06
Feat/monorepo

ovasylenko:


packages/js-plugin-aws/src/plugin.ts
```diff
@@ -1,12 +1,15 @@
...
+  MetaplexPlugin,
+} from '@metaplex-foundation/js';
 import { AwsStorageDriver } from './AwsStorageDriver';
 
 export const awsStorage = (
   client: S3Client,
   bucketName: string
 ): MetaplexPlugin => ({
-  install(metaplex: Metaplex) {
+  install(metaplex: MetaplexType) {
```
lorisleiva, line 12: Why? Is the new lint doing this?

package.json
```diff
@@ -20,26 +20,38 @@
...
+    "lint:fix": "yarn workspaces  run lint:fix",
+    "format": "prettier --write packages/ apps/",
     "docs": "typedoc",
     "deploy:docs": "yarn docs && gh-pages --dist docs --dotfiles",
     "publish": "lerna publish from-package",
     "amman:start": "DEBUG='amman:(info|error|debug)' amman start",
     "amman:stop": "amman stop",
-    "lint": "prettier -c .",
-    "lint:fix": "prettier --write ."
+    "lint": "prettier -c ."
```
lorisleiva, line 32: Should we remove this one now?

packages/js-plugin-nft-storage/src/utils.ts
```diff
@@ -58,7 +58,7 @@ export async function toEncodedCar(
 ): Promise<EncodedCar> {
   await blockstore.put(block.cid, block.bytes);
   const car = new BlockstoreCarReader([block.cid], blockstore);
-  const cid = block.cid;
+  const {cid} = block;
```
lorisleiva, line 61: Can we enforce spaces within an object?

```suggestion
  const { cid } = block;
```

packages/js/package.json
```diff
@@ -36,15 +36,15 @@
...
+		"lint:fix": "yarn eslint --fix --ext js,ts,tsx src",
     "test": "tape dist/test/**/*.test.js && yarn test:exports",
     "test:exports": "node ./test/cjs-export.test.cjs && node ./test/esm-export.test.mjs",
     "preversion": "yarn build",
     "postversion": "git push --follow-tags"
   },
   "dependencies": {
     "@bundlr-network/client": "^0.7.14",
-    "@metaplex-foundation/beet": "^0.4.0",
-    "@metaplex-foundation/beet-solana": "^0.3.0",
```
lorisleiva, line 47: Why are we removing these dependencies?

packages/js/package.json
```diff
@@ -54,8 +54,6 @@
     "bignumber.js": "^9.0.2",
     "bn.js": "^5.2.0",
     "bs58": "^5.0.0",
-    "buffer": "^6.0.3",
-    "cross-fetch": "^3.1.5",
```
lorisleiva, line 58: Same here

packages/js/src/plugins/auctionHouseModule/operations/createBid.ts
```diff
@@ -335,8 +335,8 @@ export const createBidBuilder = async (
       })
 
       // Print the Bid Receipt.
-      .when(shouldPrintReceipt, (builder) =>
-        builder.add({
+      .when(shouldPrintReceipt, (builderArg) =>
+        builderArg.add({
```
lorisleiva, line 339: Why?

packages/js/src/plugins/candyMachineModule/helpers.ts
```diff
@@ -33,7 +33,7 @@ export function parseCandyMachineItems(rawData: Buffer): CandyMachineItem[] {
 export function getCandyMachineAccountSizeFromData(data: CandyMachineData) {
   if (data.hiddenSettings != null) {
     return CONFIG_ARRAY_START;
-  } else {
+  }
```
lorisleiva, line 36: I like this change but the next part of the block needs to be indented correctly.

lorisleiva (changes requested):
Made some comments but this is more about lining than improving the monorepo structure.

What are your plans for that?

I'd like it if the linting changes could be applied separately so I can understand what's going on with the monorepo changes.

packages/js-plugin-aws/src/plugin.ts
ovasylenko: yes Metaplex is defined twice, one is shadowed by other. I renamed type as a type

packages/js/src/plugins/auctionHouseModule/operations/createBid.ts
ovasylenko: shadowing. we have builder defined upper level

packages/js/package.json
ovasylenko: It is not sued in the project. We use dep check to ensure that all deps are used and no deps are used without definition

packages/js/package.json
ovasylenko: not used/imported/etc

lorisleiva: Hmm, these are definitely used.

E.g. https://github.com/metaplex-foundation/js/blob/63c368aa9b1bbe1e86341f603e5de1bcc56eaaa9/packages/js/src/plugins/storageModule/MetaplexFile.ts#L1

`cross-fetch` should be used but it looks like it's usage was removed after a refactoring I made so I'll fix that when I have a sec.

packages/js/package.json
lorisleiva: Fair. Although the PR I'm working on at the moment makes heavy use of them. Let's remove them for now and I'll add them back on my PR. Not that `beet` is being used in the tests so it at least needs to be a dev dependency.

packages/js-plugin-aws/src/plugin.ts
lorisleiva: How is it being defined twice? I can see only one `Metaplex` instance.

https://github.com/metaplex-foundation/js/blob/f3c86ea965d5dd05ed6598c491764c1fe4194a81/packages/js-plugin-aws/src/plugin.ts

ovasylenko: disabled. reverted changes related to Metaplex type

packages/js/package.json
ovasylenko: All dev deps placed in the root package json

packages/js/package.json
ovasylenko: I think dephceck decided in favor of native node buffer. Fixed.

packages/js/src/plugins/candyMachineModule/helpers.ts
ovasylenko: not sure about it. Indentation looks ok rn

packages/js-plugin-nft-storage/src/utils.ts
ovasylenko: this one fixed with prettier

packages/js/src/plugins/candyMachineModule/helpers.ts
lorisleiva: That indentation looks off to me. See screenshot:

CleanShot 2022-09-13 at 23 28 02@2x

lorisleiva (changes requested):
Can you please only request for a review when everything has been resolved so I don't have to have a switch cost so often?

The PR's pipeline needs to pass and there should be something explaining the new lifecycle of a PR from implementation to publishing on NPM with the new infrastructure.

These linting changes make this very difficult to review.

package.json
```diff
@@ -41,20 +56,34 @@
...
+    "eslint-plugin-sort-class-members": "^1.15.2",
     "gh-pages": "^4.0.0",
-    "lerna": "^5.1.8",
-    "nx": "^14.4.3",
-    "prettier": "^2.6.0",
+    "nx": "^14.7.5",
+    "prettier": "^2.7.1",
     "rimraf": "^3.0.2",
     "rollup": "^2.72.0",
+    "rollup-plugin-generate-package-json": "^3.2.0",
```
lorisleiva, line 86: We don't need this plugin since we export using the right extension with rollup (.cjs and .mjs)

lorisleiva (changes requested):

package.json
ovasylenko: removed

packages/js-plugin-aws/package.json
```diff
@@ -36,8 +36,9 @@
     "url": "https://github.com/metaplex-foundation/js.git"
   },
   "scripts": {
-    "clean": "rimraf dist",
-    "build": "yarn clean && tsc && tsc-alias && tsc -p test/tsconfig.json && tsc-alias -p test/tsconfig.json && rollup -c",
+    "lint:test": "yarn eslint --ext js,ts,tsx src",
+    "lint:fix": "yarn eslint --fix --ext js,ts,tsx src",
+    "clean": "rimraf dist",    "build": "yarn clean && tsc && tsc-alias && tsc -p test/tsconfig.json && tsc-alias -p test/tsconfig.json && rollup -c",
```
lorisleiva, line 41:
```suggestion
    "clean": "rimraf dist",
    "build": "yarn clean && tsc && tsc-alias && tsc -p test/tsconfig.json && tsc-alias -p test/tsconfig.json && rollup -c",
```

package.json
```diff
@@ -18,21 +18,36 @@
...
     "@babel/core": "^7.17.10",
     "@babel/preset-env": "^7.17.10",
     "@babel/preset-typescript": "^7.16.7",
+    "@changesets/cli": "^2.24.4",
+    "@metaplex-foundation/amman": "^0.12.0",
+    "@metaplex-foundation/amman-client": "^0.2.2",
+    "@metaplex-foundation/beet": "^0.6.1",
+    "@metaplex-foundation/beet-solana": "^0.3.1",
+    "@nrwl/cli": "14.7.5",
+    "@nrwl/cypress": "14.7.5",
```
lorisleiva, line 44: Can you please only add the dependencies that we need? For instance, we're not using cypress so I'm not sure why we need to add this plugin on the dev dependencies. Can you please review each one and make sure it's being used otherwise remove it.

lorisleiva (changes requested):

packages/js-plugin-aws/package.json
ovasylenko: fixed

package.json
ovasylenko: checked-removed unused

package.json
```diff
@@ -36,16 +36,12 @@
     "@babel/preset-env": "^7.17.10",
     "@babel/preset-typescript": "^7.16.7",
     "@changesets/cli": "^2.24.4",
-    "@metaplex-foundation/amman": "^0.12.0",
-    "@metaplex-foundation/amman-client": "^0.2.2",
```
lorisleiva: These are used by the `amman:start` script.

packages/js-plugin-auction-house-module/package.json
```diff
@@ -0,0 +1,78 @@
+{
+  "name": "@metaplex-foundation/js-plugin-auction-house-module",
```
lorisleiva, line 2: I'm assuming this is from another PR. Please remove it from this one.

lorisleiva (changes requested):

package.json
ovasylenko: it was duplicated. it is still in the package json

packages/js-plugin-auction-house-module/package.json
ovasylenko: removed

lorisleiva (approved):

PR #275, opened 2022-09-06
Add Withdraw instruction

antey13:


packages/js/src/plugins/auctionHouseModule/operations/withdrawFromBuyerAccount.ts
```diff
@@ -0,0 +1,215 @@
...
+  WithdrawOutput
+>;
+
+/**
+ * @group Operations
+ * @category Inputs
+ */
+export type WithdrawInput = {
+  /** The Auction House from which escrow buyer withdraws funds. */
+  auctionHouse: AuctionHouse;
```
lorisleiva, line 61: minimum valuable input please

lorisleiva (changes requested):
Could you also make the CI pass?

packages/js/src/plugins/auctionHouseModule/operations/withdrawFromBuyerAccount.ts
antey13: done

packages/js/src/plugins/auctionHouseModule/operations/withdrawFromBuyerAccount.ts
```diff
@@ -0,0 +1,222 @@
...
+    | 'treasuryMint'
+    | 'feeAccountAddress'
+    >;
+  /**
+   * The buyer who withdraws funds.
+   * This expects a Signer.
+   *
+   * @defaultValue `metaplex.identity()`
+   */
+  buyer?: PublicKey | Signer;
```
lorisleiva, line 75: One empty line between parameters, please.

packages/js/src/plugins/auctionHouseModule/operations/withdrawFromBuyerAccount.ts
```diff
@@ -0,0 +1,222 @@
...
+  /**
+   * The Authority key.
+   * It is required when the buyer is not a signer.
+   *
+   * @defaultValue Defaults to not being used.
+   */
+  authority?: PublicKey | Signer;
```
lorisleiva, line 82: If it defaults to not being used, then why do we need it as a PublicKey sometimes? If `authority` is a public key, then we already have this information under `auctionHouse.authorityAddress` right?

packages/js/test/plugins/auctionHouseModule/withdrawFromBuyerAccount.test.ts
```diff
@@ -0,0 +1,70 @@
...
+  // When we withdraw 1 SOL from the buyer's escrow account.
+  await mx
+    .auctionHouse()
+    .withdrawFromBuyerAccount({
+      auctionHouse,
+      amount: sol(1),
+    })
+    .run();
+
+  // Then buyer's escrow account has minimum rent exempt SOL
```
lorisleiva, line 34: Can we also check that the receiving account has one more SOL?

packages/js/test/plugins/auctionHouseModule/withdrawFromBuyerAccount.test.ts
```diff
@@ -0,0 +1,70 @@
...
+  const buyerEscrowBalance = await mx.rpc().getBalance(buyerEscrow);
+  const minimumRentExempt = await mx.rpc().getRent(0);
+
+  t.same(
+    buyerEscrowBalance.basisPoints.toNumber(),
+    minimumRentExempt.basisPoints.toNumber()
+  );
+});
+
+test('[auctionHouseModule] it throws an error if Auctioneer Authority is not provided in Auctioneer Withdraw', async (t: Test) => {
```
lorisleiva, line 48: This is good but let's also add another test that makes sure withdrawing works when we provide an Auctioneer Auction House. This is important because we end up calling another instruction. That instruction is not tested right now.

lorisleiva (changes requested):
I've made more comments. Please make sure the linting works as well so I can merge when everything is okay. You can do that by running `yarn lint:fix`.

packages/js/src/plugins/auctionHouseModule/operations/withdrawFromBuyerAccount.ts
antey13: Right, authority can be removed here

lorisleiva: I think you still need the option to pass it as a `Signer` but not as `PublicKey | Signer` since the `PublicKey` part is redundant.

antey13:
> I've made more comments. Please make sure the linting works as well so I can merge when everything is okay. You can do that by running `yarn lint:fix`.

I do `yarn lint:fix` each time before commit, but the pipeline still fails

packages/js/src/plugins/auctionHouseModule/operations/withdrawFromBuyerAccount.ts
zaxozhu: Either authority or buyer must be a signer actually.

Here goes the rust code

```
if !wallet.to_account_info().is_signer && !authority.to_account_info().is_signer {
        return Err(AuctionHouseError::NoValidSignerPresent.into());
    }
```

packages/js/src/plugins/auctionHouseModule/operations/withdrawFromBuyerAccount.ts
```diff
@@ -0,0 +1,227 @@
...
+export const withdrawFromBuyerAccountBuilder = (
+  params: WithdrawBuilderParams,
+  metaplex: Metaplex
+): TransactionBuilder<WithdrawBuilderContext> => {
+  const { auctionHouse, auctioneerAuthority, amount } = params;
+
+  if (auctionHouse.hasAuctioneer && !params.auctioneerAuthority) {
+    throw new AuctioneerAuthorityRequiredError();
+  }
+
```
zaxozhu, line 166: I would add a check here.

What do you think? @lorisleiva @antey13 

```
if (!isSigner(seller) && !isSigner(authority)) {
  throw new RequiresSignerError();
}
```

packages/js/src/plugins/auctionHouseModule/operations/withdrawFromBuyerAccount.ts
```diff
@@ -0,0 +1,227 @@
...
+  // Signers.
+  const withdrawSigners = [buyer, authority, params.auctioneerAuthority].filter(
+    isSigner
+  );
+
+  return (
+    TransactionBuilder.make()
+
```
zaxozhu, line 219: Also, I would add 

**.setFeePayer(payer)** to allow another fee payer on a transaction similarly to other operations.

packages/js/src/plugins/auctionHouseModule/operations/withdrawFromBuyerAccount.ts
zaxozhu: So, you must do it the same way as in createListing.ts.

```
  // Update the account to be a signer since it's not covered properly by MPL due to its dynamic nature.
  const signerKeyIndex = sellInstruction.keys.findIndex((key) =>
    key.pubkey.equals(signer.publicKey)
  );
  sellInstruction.keys[signerKeyIndex].isSigner = true;
```

packages/js/src/plugins/auctionHouseModule/operations/withdrawFromBuyerAccount.ts
lorisleiva: Agreed 👌

packages/js/src/plugins/auctionHouseModule/operations/withdrawFromBuyerAccount.ts
```diff
@@ -0,0 +1,251 @@
...
+    };
+
+    withdrawInstruction = createAuctioneerWithdrawInstruction(
+      { ...accountsWithAuctioneer },
+      args
+    );
+  }
+
+  // Signers.
+  const withdrawSigners = [buyer, authority, params.auctioneerAuthority].filter(
```
zaxozhu, line 237: Try another test, create auctionHouse with different authority and try to provide this authority as a signer to this operation.

zaxozhu: You need this thing similarly to createListing to correctly apply signer. 

```
// Update the account to be a signer since it's not covered properly by MPL due to its dynamic nature.
  const signerKeyIndex = sellInstruction.keys.findIndex((key) =>
    key.pubkey.equals(signer.publicKey)
  );
  sellInstruction.keys[signerKeyIndex].isSigner = true;
```

zaxozhu: Currently, it works because fee payer equals authority.

zaxozhu (approved):

packages/js/src/plugins/auctionHouseModule/AuctionHouseBuildersClient.ts
```diff
@@ -49,6 +51,11 @@ export class AuctionHouseBuildersClient {
     return depositToBuyerAccountBuilder(this.metaplex, input);
   }
 
+  /** {@inheritDoc withdrawFromBuyerAccountBuilder} */
+  withdrawFromBuyerAccount(input: WithdrawFromBuyerAccountBuilderParams) {
+    return withdrawFromBuyerAccountBuilder(this.metaplex, input);
+  }
```
lorisleiva, line 57: These should be ordered alphabetically (see comment in other PR).

packages/js/src/plugins/auctionHouseModule/errors.ts
```diff
@@ -127,3 +127,17 @@ export class CreateListingRequiresSignerError extends AuctionHouseError {
     });
   }
 }
+
+/** @group Errors */
+export class WithdrawRequiresSignerError extends AuctionHouseError {
```
lorisleiva, line 132: Since this error is just for withdrawing from the buyer's account its name should be more explicit.

```suggestion
export class WithdrawFromBuyerAccountRequiresSignerError extends AuctionHouseError {
```

packages/js/src/plugins/auctionHouseModule/operations/withdrawFromBuyerAccount.ts
```diff
@@ -70,19 +73,30 @@ export type WithdrawInput = {
...
 
   /**
    * The Authority key.
    * It is required when the buyer is not a signer.
+   * There must be one and only one signer; Authority or Buyer must sign.
    *
    * @defaultValue Defaults to not being used.
    */
-  authority?: Signer;
+  authority?: PublicKey | Signer;
```
lorisleiva, line 99: Again, this does not need to be a `PublicKey`. We still potentially require this as a Signer since either the authority or the buyer needs to sign the transaction but we already know the authority as a PublicKey from the `auctionHouse.authorityAddress` parameter.

It should be:

```ts
authority?: Signer;
```

lorisleiva (changes requested):

lorisleiva (changes requested):
One last comment

packages/js/src/plugins/auctionHouseModule/plugin.ts
```diff
@@ -52,6 +52,10 @@ import {
   updateAuctionHouseOperationHandler,
   withdrawFromBuyerAccountOperation,
   withdrawOperationHandler,
```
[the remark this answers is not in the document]
antey13: done

packages/js/src/plugins/auctionHouseModule/operations/withdrawFromBuyerAccount.ts
```diff
@@ -156,7 +160,10 @@ export type WithdrawFromBuyerAccountBuilderParams = Omit<
  * @group Transaction Builders
  * @category Contexts
  */
-export type WithdrawBuilderContext = Omit<WithdrawOutput, 'response'>;
+export type WithdrawBuilderContext = Omit<
```
lorisleiva, line 163: You missed one. 😬😅

lorisleiva (changes requested):

packages/js/src/plugins/auctionHouseModule/AuctionHouseClient.ts
```diff
@@ -147,6 +150,15 @@ export class AuctionHouseClient {
...
 
+  /** {@inheritDoc withdrawFromBuyerAccountOperation} */
+  withdrawFromBuyerAccount(
+    input: WithdrawFromBuyerAccountInput
+  ): Task<WithdrawFromBuyerAccountOutput> {
+    return this.metaplex
+      .operations()
+      .getTask(withdrawFromBuyerAccountOperation(input));
+  }
+
```
lorisleiva, line 161: Sorry to be a pain, but could you also order that method alphabetically?

packages/js/src/plugins/auctionHouseModule/operations/withdrawFromBuyerAccount.ts
```diff
@@ -0,0 +1,266 @@
...
+    );
+
+    const accountsWithAuctioneer = {
+      ...accounts,
+      auctioneerAuthority: auctioneerAuthority.publicKey,
+      ahAuctioneerPda,
+    };
+
+    withdrawInstruction = createAuctioneerWithdrawInstruction(
+      { ...accountsWithAuctioneer },
```
lorisleiva, line 241:
```suggestion
      accountsWithAuctioneer,
```

lorisleiva (changes requested):

lorisleiva (approved):

PR #276, opened 2022-09-07
Implement Withdraw from Fees & Treasury

zaxozhu:


packages/js/src/plugins/auctionHouseModule/operations/withdrawFromFeeAccount.ts
```diff
@@ -0,0 +1,190 @@
...
+ * @group Transaction Builders
+ * @category Contexts
+ */
+export type WithdrawFromFeeAccountBuilderContext = Omit<
+  WithdrawFromFeeAccountOutput,
+  'response'
+>;
+
+/**
+ * Adds funds to the user's buyer escrow account for the given auction house.
```
lorisleiva, line 136: This comment needs updating.

packages/js/src/plugins/auctionHouseModule/operations/withdrawFromTreasuryAccount.ts
```diff
@@ -0,0 +1,198 @@
...
+ * @group Transaction Builders
+ * @category Contexts
+ */
+export type WithdrawFromTreasuryAccountBuilderContext = Omit<
+  WithdrawFromTreasuryAccountOutput,
+  'response'
+>;
+
+/**
+ * Adds funds to the user's buyer escrow account for the given auction house.
```
lorisleiva, line 138: This comment needs updating.

packages/js/test/plugins/auctionHouseModule/withdrawFromFeeAccount.test.ts
```diff
@@ -0,0 +1,73 @@
...
+  // When we withdraw 1 SOL from fee account.
+  await mx
+    .auctionHouse()
+    .withdrawFromFeeAccount({
+      auctionHouse,
+      amount: sol(1),
+    })
+    .run();
+
+  // Then fee account has 99 SOL in it.
```
lorisleiva, line 29: Can we also check that the destination has one more SOL? I know we check this on the other test but it would be nice to have one test that checks both the sending and the receiving part to be safe.

packages/js/test/plugins/auctionHouseModule/withdrawFromTreasuryAccount.test.ts
```diff
@@ -0,0 +1,73 @@
...
+  killStuckProcess,
+  createNft,
+  createWallet,
+} from '../../helpers';
+import { createAuctionHouse } from './helpers';
+
+killStuckProcess();
+
+test('[auctionHouseModule] withdraw from treasury account on an Auction House', async (t: Test) => {
+  // Given we have an Auction House with fee that equals 100 basis points and an NFT.
```
lorisleiva, line 14:
```suggestion
  // Given we have an Auction House with fee that equals 10% and an NFT.
```

According to the code, that's `1000` basis points not `100`. Also, in the comments, let's make it percentages so its clearer.

packages/js/test/plugins/auctionHouseModule/withdrawFromTreasuryAccount.test.ts
```diff
@@ -0,0 +1,73 @@
...
+  const mx = await metaplex();
+  const buyer = await createWallet(mx);
+  const nft = await createNft(mx);
+
+  const { auctionHouse } = await createAuctionHouse(mx, null, {
+    sellerFeeBasisPoints: 1000,
+  });
+
+  // And we deposited 2 SOL to the buyer's escrow, so 0.1 SOL fee can be withdrawn.
```
lorisleiva, line 23: With 10% fees, we should be able to take 0.2 SOL. Could you update the rest of the test accordingly?

```suggestion
  // And we deposited 2 SOL to the buyer's escrow, so 0.2 SOL fee can be withdrawn.
```

Note that the fee is taken from the seller. So the buyer doesn't have to have more SOLs to pay for the fees. The buy's deposit account and the listing price can be exactly the same amount.

packages/js/test/plugins/auctionHouseModule/withdrawFromTreasuryAccount.test.ts
```diff
@@ -0,0 +1,73 @@
...
+  metaplex,
+  killStuckProcess,
+  createNft,
+  createWallet,
+} from '../../helpers';
+import { createAuctionHouse } from './helpers';
+
+killStuckProcess();
+
+test('[auctionHouseModule] withdraw from treasury account on an Auction House', async (t: Test) => {
```
lorisleiva, line 13: A note on that test.

I think it's great that it is testing the royalty system but I think 80% of this test belongs to the "execute sale" operation.

Here we should get straight to the point and airdrop the treasury account so we can test the withdraw logic (just like we did for the `withdrawFromFeeAccount` operation).

lorisleiva (changes requested):
Overall, that's a great PR, thanks!

packages/js/test/plugins/auctionHouseModule/withdrawFromFeeAccount.test.ts
```diff
@@ -35,6 +39,19 @@ test('[auctionHouseModule] withdraw from fee account on an Auction House', async
...
+  const feeWithdrawalDestinationBalance = await mx
+    .rpc()
+    .getBalance(toPublicKey(auctionHouse.feeWithdrawalDestinationAddress));
+
+  t.same(
+    subtractAmounts(
+      feeWithdrawalDestinationBalance,
+      sol(100)
+    ).basisPoints.toNumber(),
+    sol(1).basisPoints.toNumber()
```
lorisleiva, line 53: Can we just check that it's `101`?

Also, let's make it explicit that the `feeWithdrawalDestinationAddress` initially has 100 SOL in the "Given" part of the test.

packages/js/test/plugins/auctionHouseModule/withdrawFromTreasuryAccount.test.ts
```diff
@@ -1,73 +1,50 @@
...
+  const feeWithdrawalDestinationBalance = await mx
+    .rpc()
+    .getBalance(toPublicKey(auctionHouse.treasuryWithdrawalDestinationAddress));
+
+  t.same(
+    subtractAmounts(
+      feeWithdrawalDestinationBalance,
+      sol(100)
+    ).basisPoints.toNumber(),
+    sol(1).basisPoints.toNumber()
```
lorisleiva, line 48: Same comment here.

lorisleiva (changes requested):

lorisleiva (approved):
That's great, thanks!

PR #278, opened 2022-09-08
docs: fixed links

tjkyner:
Fixing a small pathing issue that was causing links to the `createNft.ts` file to lead to a 404 page.


lorisleiva:
Thanks! And thanks for updating both readme

Issue #234, opened 2022-08-15
Candy Machine: Zero-decimal SPL Tokens


cosimo-rip:
Hi there! I've been working with the candy machine plugin, and in testing, believe I've come across an issue with SPL token-based mints.

I think the library does not handle SPL tokens with varying decimals. In this instance, I have an SPL token with zero decimals, and when setting a price of 1, the resulting candy machine (through create or update methods) ends up deployed with a resulting price of `1,000,000,000` (off by a factor of 10^9).

At first I was working with the JSON based methods like `createFromJsonConfig`. Since price is just a `number` there I was simply providing `price: 1`.

I assumed I might need to move to the lower level `create` method for more control, and tried to solve this by doing:
`price: token(myPrice, 0, "CUSTOM")`

But alas, in both cases price came out as  `1,000,000,000`.

Checked out the tests for guidance and noticed that SPL tokens are always generated with `mx.tokens().createTokenWithMint()`, so I'm wondering if this is just an unexplored case.
https://github.com/metaplex-foundation/js/blob/8d2183d84f8ce4e6f383b8578f84df6cdc5d1049/packages/js/test/plugins/candyMachineModule/createCandyMachine.test.ts#L130

I also see here that the price is assumed to always be compatible with `lamports()`:
https://github.com/metaplex-foundation/js/blob/8d2183d84f8ce4e6f383b8578f84df6cdc5d1049/packages/js/src/plugins/candyMachineModule/CandyMachine.ts#L139

Please let me know if I'm doing something wrong! (Also, FWIW, Sugar CLI works fine with zero-decimal tokens).

lorisleiva:
Thanks again for your contribution, I will publish this with the next breaking change version since the `main` branch already contains breaking changes. Hopefully, that should be this week. 🤞

lorisleiva:
Version `0.16.0` contains your changes. 🎉

cosimo-rip:
perfect timing! skullana fest thanks you 😉

PR #279, opened 2022-09-10
Candy Machine: accurate Amount if a `tokenMintAddress` is specified, fixes [#234]

cosimo-rip:
This is a fix for #234 to handle the amounts for Candy Machine correctly when supplying arbitrary decimal SPL token treasuries.

### DIFF
--- a/packages/js/src/plugins/candyMachineModule/operations/findCandyMachinesByPublicKeyField.ts
+++ b/packages/js/src/plugins/candyMachineModule/operations/findCandyMachinesByPublicKeyField.ts
@@ -1,9 +1,11 @@
 import { UnreachableCaseError } from '@/errors';
 import { Metaplex } from '@/Metaplex';
+import { Mint, toMint, toMintAccount } from '@/plugins/tokenModule';
 import { Operation, OperationHandler, useOperation } from '@/types';
 import { DisposableScope, zipMap } from '@/utils';
 import { Commitment, PublicKey } from '@solana/web3.js';
 import {
+  CandyMachineAccount,
   parseCandyMachineAccount,
   parseCandyMachineCollectionAccount,
 } from '../accounts';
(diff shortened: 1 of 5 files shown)


packages/js/src/plugins/candyMachineModule/operations/findCandyMachinesByPublicKeyField.ts
```diff
@@ -112,17 +112,31 @@ export const findCandyMachinesByPublicKeyFieldOperationHandler: OperationHandler
...
             ? parseCandyMachineCollectionAccount(unparsedCollectionAccount)
             : null;
 
-          return toCandyMachine(account, unparsedAccount, collectionAccount);
+          const mint = account.data.tokenMint
+            ? await metaplex
+                .tokens()
+                .findMintByAddress({ address: account.data.tokenMint })
+                .run()
+            : null;
```
lorisleiva, line 129: Sadly, this is not performant as it will create an additional blockchain request for every SPL Token Candy Machine in the list. We should fetch all mint accounts at once using a `getMultipleAccounts` call and then reconciliate each mint account with its candy machine account.

lorisleiva (changes requested):
Hi there 👋 

Thanks for this contribution. You've ticked all the boxes on this PR but, unfortunately, the reason why this wasn't already done is that it needs to be done in a more complex way to be performant. See my comment below.

packages/js/src/plugins/candyMachineModule/operations/findCandyMachinesByPublicKeyField.ts
cosimo-rip: @lorisleiva thanks for the feedback Loris. I have a better understanding of what you were doing in that part of the codebase now.

I took another shot at the implementation and added an add'l test for parsing out the token mint. Let me know if this is more in line w/what you wanted

packages/js/src/plugins/candyMachineModule/operations/findCandyMachinesByPublicKeyField.ts
```diff
@@ -121,8 +122,45 @@ export const findCandyMachinesByPublicKeyFieldOperationHandler: OperationHandler
...
+            account,
+            unparsedAccount,
+            collectionAccount,
+          };
         }
       );
+
+      // Find mint details for all unique SPL tokens used
+      // in candy machines that have non-null `tokenMint`
+      const tokenMints = new Set(
```
lorisleiva, line 135: Just curious: Does `Set` works with the `PublicKey` class? How does it know how to identify duplicate values? Is it doing it by reference or is there a magic method that `Set` picks up on?

packages/js/src/plugins/candyMachineModule/operations/findCandyMachinesByPublicKeyField.ts
```diff
@@ -121,8 +122,45 @@ export const findCandyMachinesByPublicKeyFieldOperationHandler: OperationHandler
...
+
+      // Then combine to return all `CandyMachine` results
+      return candyMachinesDetails.map((details) => {
+        const tokenMint = details.account.data.tokenMint;
+        return toCandyMachine(
+          details.account,
+          details.unparsedAccount,
+          details.collectionAccount,
+          tokenMint
+            ? mints.find((mint) => mint.address === tokenMint) ?? null
```
lorisleiva, line 161: You're comparing public keys by reference here. So the same public keys could still return false in this comparison. You need to use `mint.address.equals(tokenMint)`. Even better, instead of having a list of `Mint`, you could have a map `Record<string, Mint>` so finding a Mint is O(1).

packages/js/src/plugins/candyMachineModule/operations/findCandyMachinesByPublicKeyField.ts
```diff
@@ -121,8 +122,45 @@ export const findCandyMachinesByPublicKeyFieldOperationHandler: OperationHandler
...
+          .map((details) => details.account.data.tokenMint)
+          .filter((tokenMint): tokenMint is PublicKey => tokenMint !== null)
+      );
+
+      let mints: Mint[] = [];
+
+      if (tokenMints.size > 0) {
+        const unparsedMintAccounts = await metaplex
+          .rpc()
+          .getMultipleAccounts([...tokenMints], commitment);
```
lorisleiva, line 146: If you're up for the challenge, we could combine this `getMultipleAccounts` call with the one that fetches collections since they can both be done in parallel. We'd just need to slice the returned accounts so we know which ones are collections and which ones are mint.

lorisleiva (changes requested):
Thanks for that! I've made a few comments to make this even better. Let me know what you think! 😊

packages/js/src/plugins/candyMachineModule/operations/findCandyMachinesByPublicKeyField.ts
cosimo-rip: I intended to map to string there - good catch.

packages/js/src/plugins/candyMachineModule/operations/findCandyMachinesByPublicKeyField.ts
cosimo-rip: Sure!

cosimo-rip:
Thanks for the reviews and suggestions. Just made some changes to combine the `getMultipleAccounts` calls, move to `Record`, and fixed the other issues you caught. Also improved the test to have more coverage of the code in the PR w/a third candy machine that has redundant token mints.

lorisleiva:
@cosimo-rip That's awesome thank you! I'm going to be away for a few days but I'll check this out thoroughly when I'm back.

packages/js/test/plugins/candyMachineModule/findCandyMachinesByPublicKeyField.test.ts
```diff
@@ -66,3 +67,60 @@ test('[candyMachineModule] find all candy machines by authority', async (t) => {
     );
   });
 });
+
+test('[candyMachineModule] find all candy machines correctly parses token mints', async (t) => {
```
lorisleiva, line 71: This is a great test thanks!

Could we just make the following adjustment to also make sure the collection PDAs are being fetched and parsed properly:

Could the first two candy machines each have their own collection set using the `collection` parameter? You may create them using the helper method `createCollectionNft`.

Then we can ensure they all have the right collection on the results. Feel free to test each candy machine one by one instead of doing a loop to make this easier.

lorisleiva (changes requested):
This is perfect, exactly what I had in mind! Thank you so much for that!

I've just added a little comment to improve one of the tests and then let's get this merged. 🙂

packages/js/test/plugins/candyMachineModule/findCandyMachinesByPublicKeyField.test.ts
cosimo-rip: @lorisleiva done!

lorisleiva (approved):
Thank you! 🍺

PR #283, opened 2022-09-13
Add operations builders to AuctionHouseBuildersClient

antey13:


zaxozhu (approved):

lorisleiva (changes requested):
Thanks! Can we please keep these methods ordered alphabetically?

This is because these methods are just registering transaction builders so it removes the need to think in what order we want them when we register them here.

packages/js/src/plugins/auctionHouseModule/AuctionHouseBuildersClient.ts
```diff
@@ -41,6 +45,15 @@ export class AuctionHouseBuildersClient {
     return createBidBuilder(this.metaplex, input);
   }
 
+  /** {@inheritDoc cancelBidBuilder} */
+  cancelBid(input: CancelBidBuilderParams) {
+    return cancelBidBuilder(input);
+  }
+  /** {@inheritDoc cancelListingBuilder} */
```
zaxozhu, line 52: New line here, please

antey13: done

lorisleiva:
Thanks!

PR #289, opened 2022-09-15
Add Direct Buy & Sell operations

antey13:


packages/js/test/plugins/auctionHouseModule/buy.test.ts
```diff
@@ -0,0 +1,68 @@
...
+  const { listing } = await mx
+    .auctionHouse()
+    .list({
+      auctionHouse,
+      mintAccount: nft.address,
+      price: sol(1),
+    })
+    .run();
+
+  // When we execute a sale with given listing and created private bid on that NFT for 1 SOL.
```
zaxozhu, line 35:
```suggestion
  // When we execute direct buy with the given listing.
```

zaxozhu (approved):

packages/js/src/plugins/auctionHouseModule/AuctionHouseClient.ts
```diff
@@ -236,6 +249,11 @@ export class AuctionHouseClient {
       .getTask(getBuyerBalanceOperation(options));
   }
 
+  /** {@inheritDoc saleOperation} */
+  sell(options: SellInput): Task<SellOutput> {
+    return this.metaplex.operations().getTask(sellOperation(options));
+  }
+
```
lorisleiva, line 256: `sell` as a method name is great because it's in the context of the auction house client.

However, `SellOperation`, `SellInput`, etc. needs to be more specific because they are exported globally (without any context).

The same goes for the `buy` operation.

May I suggest:
- `AuctionHouseDirectSellOperation`
- `AuctionHouseDirectBuyOperation`

Please update everything accordingly: transaction builder, inputs, outputs, file names, etc. Just keep the methods `sell` and `buy` as-is.

packages/js/src/plugins/auctionHouseModule/operations/buy.ts
```diff
@@ -0,0 +1,214 @@
...
+ */
+export type BuyOperation = Operation<typeof Key, BuyInput, BuyOutput>;
+
+/**
+ * @group Operations
+ * @category Inputs
+ */
+export type BuyInput = {
+  /** The Auction House in which to create a Bid and execute a Sale. */
+  auctionHouse: AuctionHouse;
```
lorisleiva, line 49: Minimum Valuable Input!

packages/js/src/plugins/auctionHouseModule/operations/buy.ts
```diff
@@ -0,0 +1,214 @@
...
+    return await metaplex
+      .auctionHouse()
+      .executeSale({
+        bid,
+        listing,
+        ...rest,
+      })
+      .run();
+  },
+};
```
lorisleiva, line 214: This is not what I expected.

The direct sell/buy operations should still expose a transaction builder that delegates to the other two transaction builders. Then, this operation handler simply uses it like most other handlers.

packages/js/src/plugins/auctionHouseModule/operations/buy.ts
```diff
@@ -0,0 +1,214 @@
...
+
+  /** The number of tokens bought. */
+  tokens: SplTokenAmount;
+
+  /** A model that keeps information about the Purchase. */
+  purchase: Purchase;
+
+  /** The blockchain response from sending and confirming the transaction. */
+  response: SendAndConfirmTransactionResponse;
+};
```
lorisleiva, line 180: The output should return the purchase and the bid/listing within the output object. It should not be inlined like this.

packages/js/src/plugins/auctionHouseModule/operations/buy.ts
```diff
@@ -0,0 +1,214 @@
...
+   * @defaultValue No default value.
+   */
+  tokenAccount?: Option<PublicKey>;
+
+  /**
+   * The buyer's price.
+   *
+   * @defaultValue 0 SOLs or tokens.
+   */
+  price?: SolAmount | SplTokenAmount;
```
lorisleiva, line 106: The price should default to the listing's price.

packages/js/src/plugins/auctionHouseModule/operations/buy.ts
```diff
@@ -0,0 +1,214 @@
...
+   * The number of tokens to bid for.
+   * For an NFT bid it must be 1 token.
+   *
+   * When a Fungible Asset is put on sale.
+   * The buyer can then create a buy order of said assets that is
+   * less than the token_size of the sell order.
+   *
+   * @defaultValue 1 token.
+   */
+  tokens?: SplTokenAmount;
```
lorisleiva, line 118: The tokens should default to the listing's tokens.

packages/js/src/plugins/auctionHouseModule/operations/buy.ts
```diff
@@ -0,0 +1,214 @@
...
+   * Prints the bid receipt.
+   * The receipt holds information about the bid,
+   * So it's important to print it if you want to use the `Bid` model
+   *
+   * The receipt printing is skipped for the Auctioneer Auction House
+   * Since it currently doesn't support it.
+   *
+   * @defaultValue `true`
+   */
+  printReceipt?: boolean;
```
lorisleiva, line 130: The comment should explain that this will be used for both printing the bid and the purchase receipts.

packages/js/src/plugins/auctionHouseModule/operations/buy.ts
```diff
@@ -0,0 +1,214 @@
...
+   */
+  auctioneerAuthority?: Signer;
+
+  /**
+   * The token account address that's associated to the asset a bid created is for.
+   * If this or seller isn't provided in the Listing, then the bid will be public.
+   *
+   * @defaultValue No default value.
+   */
+  tokenAccount?: Option<PublicKey>;
```
lorisleiva, line 99: This is not necessary since we're buying the listing's asset directly. It will always be a private bid and we know the token account via the `listing` parameter.

packages/js/test/plugins/auctionHouseModule/buy.test.ts
```diff
@@ -0,0 +1,68 @@
...
+  const { listing } = await mx
+    .auctionHouse()
+    .list({
+      auctionHouse,
+      mintAccount: nft.address,
+      price: sol(1),
+    })
+    .run();
+
+  // When we execute direct buy with the given listing.
```
lorisleiva, line 35: We need more checks here on the state of every account. Buyer, Seller and Purchase.

packages/js/test/plugins/auctionHouseModule/buy.test.ts
```diff
@@ -0,0 +1,68 @@
...
+  metaplex,
+  spokSameAmount,
+  spokSamePubkey,
+} from '../../helpers';
+import { createAuctionHouse } from './helpers';
+import { findAssociatedTokenAccountPda, Purchase } from '@/index';
+
+killStuckProcess();
+
+test('[auctionHouseModule] buy on an Auction House', async (t: Test) => {
```
lorisleiva, line 17: We need more tests (for both selling and buying):
- One with minimum input: allowing all default values to do their job and asserting it works.
- One with maximum input: making sure the operation still works when we provide explicit values for everything.
- One that asserts the operation fails is the input is incorrect (e.g. wrong price).
- When buying, one that checks partial buys as soon as the other PR is merged.

lorisleiva (changes requested):

packages/js/src/plugins/auctionHouseModule/AuctionHouseClient.ts
antey13: Maybe just DirectSellOperation would be enough? As for me, it is obvious from here that it is AuctionHouse operations

lorisleiva: Okay let's go with `DirectSell` and `DirectBuy` 👌

packages/js/src/plugins/auctionHouseModule/operations/buy.ts
lorisleiva: Why do we need the full object for `executeSale`?

lorisleiva: Ignore me, I know why. We need it to recreate the LazyPurchase.

antey13: I am not sure, but it is required by `ExecuteSaleInput` to have `auctionHouse: AuctionHouse;`

packages/js/src/plugins/auctionHouseModule/operations/directBuy.ts
```diff
@@ -0,0 +1,299 @@
...
+    metadataAddress: createBidBuilderContext.metadata,
+    receiptAddress: createBidBuilderContext.receipt,
+    purchaseReceiptAddress: null,
+    isPublic: Boolean(createBidBuilderContext.tokenAccount),
+    price: listing.price,
+    tokens: createBidBuilderContext.tokens.basisPoints,
+    createdAt: now(),
+    canceledAt: null,
+  };
+
```
zaxozhu, line 260:
```suggestion
const bid =  metaplex
        .auctionHouse()
        .loadBid({ lazyBid })
```

packages/js/src/plugins/auctionHouseModule/operations/directBuy.ts
```diff
@@ -0,0 +1,299 @@
...
+        tokens: createBidBuilderContext.tokens.basisPoints,
+        tradeStateAddress: createBidBuilderContext.buyerTradeState,
+      } as Bid,
+      listing,
+      ...rest,
+    });
+
+  return TransactionBuilder.make<DirectBuyBuilderContext>()
+    .setContext({
+      bid: lazyBid as Bid,
```
zaxozhu, line 282:
```suggestion
      bid
```

packages/js/src/plugins/auctionHouseModule/operations/directBuy.ts
```diff
@@ -0,0 +1,251 @@
...
+ * @group Operations
+ * @category Handlers
+ */
+export const directBuyOperationHandler: OperationHandler<DirectBuyOperation> = {
+  handle: async (operation: DirectBuyOperation, metaplex: Metaplex) => {
+    const { auctionHouse } = operation.input;
+    const { receipt, bid, response } = await directBuyBuilder(
+      metaplex,
+      operation.input
+    ).then((buyBuilder) =>
```
zaxozhu, line 136: You have **await** above, you don't need then here.

zaxozhu: `await buyBuilder.sendAndConfirm(metaplex, operation.input.confirmOptions)`

packages/js/src/plugins/auctionHouseModule/operations/directBuy.ts
```diff
@@ -0,0 +1,251 @@
...
+      listing,
+      printReceipt,
+      bookkeeper,
+      ...rest,
+    });
+
+  return TransactionBuilder.make<DirectBuyBuilderContext>()
+    .setContext({
+      bid,
+      receipt: saleBuilder.getContext().receipt as Pda,
```
zaxozhu, line 247:
```suggestion
      receipt: saleBuilder.getContext().receipt,
```

packages/js/src/plugins/auctionHouseModule/operations/directBuy.ts
```diff
@@ -0,0 +1,251 @@
...
+export type DirectBuyBuilderParams = Omit<DirectBuyInput, 'confirmOptions'>;
+
+/**
+ * @group Transaction Builders
+ * @category Contexts
+ */
+export type DirectBuyBuilderContext = Omit<
+  DirectBuyOutput,
+  'response' | 'purchase'
+> & { receipt: Pda };
```
zaxozhu, line 169: It's optional

```suggestion
> & { receipt: Option<Pda> };
```

packages/js/src/plugins/auctionHouseModule/operations/directBuy.ts
```diff
@@ -0,0 +1,251 @@
...
+  const {
+    auctionHouse,
+    listing,
+    buyer = metaplex.identity(),
+    authority = auctionHouse.authorityAddress,
+    bookkeeper = metaplex.identity(),
+    ...rest
+  } = params;
+  const { tokens, price, asset, sellerAddress } = listing;
+  const printReceipt = true;
```
zaxozhu, line 198: This should depend on listing.receipt

packages/js/src/plugins/auctionHouseModule/operations/directBuy.ts
```diff
@@ -0,0 +1,251 @@
...
+  handle: async (operation: DirectBuyOperation, metaplex: Metaplex) => {
+    const { auctionHouse } = operation.input;
+    const { receipt, bid, response } = await directBuyBuilder(
+      metaplex,
+      operation.input
+    ).then((buyBuilder) =>
+      buyBuilder.sendAndConfirm(metaplex, operation.input.confirmOptions)
+    );
+
+    const purchase = await metaplex
```
zaxozhu, line 140: It won't work if the listing doesn't have a receipt, you can check executeSale.

const shouldPrintReceipt =
    (params.printReceipt ?? true) &&
    Boolean(**listing.receiptAddress && bid.receiptAddress** && !isPartialSale);

packages/js/src/plugins/auctionHouseModule/operations/directSell.ts
```diff
@@ -0,0 +1,275 @@
...
+ * @category Handlers
+ */
+export const directSellOperationHandler: OperationHandler<DirectSellOperation> =
+  {
+    handle: async (operation: DirectSellOperation, metaplex: Metaplex) => {
+      const { auctionHouse } = operation.input;
+      const { receipt, listing, response } = await directSellBuilder(
+        metaplex,
+        operation.input
+      ).then((buyBuilder) =>
```
zaxozhu, line 146: Same as above, please.

packages/js/src/plugins/auctionHouseModule/plugin.ts
```diff
@@ -4,6 +4,8 @@ import type { ErrorWithLogs, MetaplexPlugin } from '@/types';
 import { AuctionHouseClient } from './AuctionHouseClient';
 import { AuctionHouseProgram } from './program';
 import {
+  directBuyOperation,
+  directBuyOperationHandler,
```
lorisleiva, line 8: VS Code > Organize Imports

packages/js/test/plugins/auctionHouseModule/directBuy.test.ts
```diff
@@ -0,0 +1,147 @@
...
+    })
+    .run();
+
+  // When we execute direct buy with the given listing.
+  const { purchase } = await mx
+    .auctionHouse()
+    .buy({
+      auctionHouse,
+      listing,
+      buyer,
```
lorisleiva, line 43: "Minimum inputs" means we should not provide the `buyer` which defaults to the identity. Instead, the `seller` should be explicitly defined so they are different wallets.

packages/js/test/plugins/auctionHouseModule/directBuy.test.ts
```diff
@@ -0,0 +1,147 @@
...
+
+  // Then we expect an error.
+  await assertThrows(
+    t,
+    promise,
+    /You are trying to execute a direct buy, but direct buy are not supported in Auctioneer./
+  );
+});
+
+test('[auctionHouseModule] buy on an Auction House', async (t: Test) => {
```
lorisleiva, line 110: Is this the maximum input version? If so, it should be explicit and the test should be right after the minimum input version. I think it might need more parameters as well and all provided parameters should be checked on the "Then" part of the test.

packages/js/test/plugins/auctionHouseModule/directSell.test.ts
```diff
@@ -0,0 +1,145 @@
...
+  spokSamePubkey,
+} from '../../helpers';
+import { createAuctionHouse } from './helpers';
+import { findAssociatedTokenAccountPda, Purchase } from '@/plugins';
+import spok, { Specifications } from 'spok';
+import { Keypair } from '@solana/web3.js';
+
+killStuckProcess();
+
+test('[auctionHouseModule] instant sale on an Auction House with minimum input', async (t: Test) => {
```
lorisleiva, line 19: See comments for direct buy here too.

lorisleiva (changes requested):

packages/js/src/plugins/auctionHouseModule/operations/directSell.ts
```diff
@@ -0,0 +1,307 @@
...
+    // Load asset token if the bid was public and there is no token data in the asset model.
+    const asssetTokenAddress = findAssociatedTokenAccountPda(
+      asset.address,
+      toPublicKey(seller)
+    );
+
+    listingAsset = await metaplex
+      .nfts()
+      .findByToken({ token: asssetTokenAddress })
+      .run();
```
lorisleiva, line 255: @zaxozhu I want to avoid fetching from the blockchain at the transaction builder level.

Can we do that check on the Operation Handler and accept a non-optional `asset` parameter for the transaction builder?

packages/js/src/plugins/auctionHouseModule/operations/directSell.ts
```diff
@@ -0,0 +1,307 @@
...
+  let listingAsset: NftWithToken | SftWithToken;
+
+  if (isNftWithToken(asset) || isSftWithToken(asset)) {
+    listingAsset = asset;
+  } else {
+    // Load asset token if the bid was public and there is no token data in the asset model.
+    const asssetTokenAddress = findAssociatedTokenAccountPda(
+      asset.address,
+      toPublicKey(seller)
+    );
```
lorisleiva, line 250: @zaxozhu Can you remind me why `tokenAccount` uses the buyer address whereas `asssetTokenAddress` uses the seller address?

packages/js/test/plugins/auctionHouseModule/directSell.test.ts
```diff
@@ -0,0 +1,160 @@
...
+
+test('[auctionHouseModule] sell on an Auction House with maximum input', async (t: Test) => {
+  // Given we have an Auction House and an NFT.
+  const mx = await metaplex();
+  const buyer = await createWallet(mx);
+  const seller = await createWallet(mx);
+
+  const nft = await createNft(mx, { tokenOwner: seller.publicKey });
+  const auctionHouse = await createAuctionHouse(mx, null, {
+    authority: seller,
```
lorisleiva, line 108: @zaxozhu When you create tests for "maximum input" scenarios, please create a new Keypair for every single parameter unless they have to be the same.

Here, there should be another Keypair for the authority of the auction house.

Also use `Keypair.generate()` instead of `createWallet` whenever you can so that:
- A. the tests are a bit faster
- B. we make sure that no lamports are required from the former otherwise an error will be thrown.

packages/js/src/plugins/auctionHouseModule/operations/directSell.ts
zaxozhu: Oh, yeah, that looks off. There must be a seller's token account.

packages/js/src/plugins/auctionHouseModule/models/Bid.ts
```diff
@@ -17,84 +17,97 @@ import { AuctionHouse } from './AuctionHouse';
 import { Nft, NftWithToken, Sft, SftWithToken } from '../../nftModule';
 
 /** @group Models */
-export type Bid = Readonly<
-  {
-    /** A model identifier to distinguish models in the SDK. */
-    model: 'bid';
+export type Bid = Readonly<PublicBid | PrivateBid>;
```
lorisleiva, line 20: Can you revert that change? This will not play nicely with our typedocs.

If you want explicit types for `PublicBid` and `PrivateBid`, could you do something like this?

```ts
export type PublicBid = Bid & { isPublic: true };
export type PrivateBid = Bid & { isPublic: false };
```

packages/js/src/plugins/auctionHouseModule/operations/createListing.ts
```diff
@@ -407,6 +407,7 @@ export const createListingBuilder = (
     key.pubkey.equals(signer.publicKey)
   );
   sellInstruction.keys[signerKeyIndex].isSigner = true;
+  sellInstruction.keys[signerKeyIndex].isWritable = true;
```
lorisleiva, line 410: Are you sure it needs to be writable?

packages/js/src/plugins/auctionHouseModule/operations/directSell.ts
```diff
@@ -90,7 +84,10 @@ export type DirectSellInput = {
    *
    * This includes, its asset, auction house address, buyer, receipt address etc.
    */
-  bid: Omit<Bid, 'bookkeeperAddress' | 'purchaseReceiptAddress' | 'createdAt'>;
+  bid: Omit<
+    PrivateBid,
```
lorisleiva, line 88: I still think we should allow "direct sell" to work on public bids. It's just that we need to ask the seller to provide the token account and default to the ATA.

To Recap:
- We accept a `bid` which can bid public or private and an optional `sellerTokenAccount` which defaults to the ATA between the seller and `bid.asset.address`.
- If `bid` is private, then we need to make sure the token accounts match.
- If `bid` is public, then we use the seller token account to figure out which asset to sell.
- Since we need to return the bid from the operation, we'll need to fetch the asset but we need to do that on the operation handler. Not on the transaction builder.

packages/js/test/plugins/auctionHouseModule/directBuy.test.ts
```diff
@@ -15,112 +15,118 @@ import { Keypair } from '@solana/web3.js';
...
-      listing,
-      buyer,
-      price: sol(1),
-    })
-    .run();
-
-  // Then we created and returned the new Purchase
-  t.equal(purchase.asset.address.toBase58(), nft.address.toBase58());
-});
+// test('[auctionHouseModule] buy on an Auction House with minimum input', async (t: Test) => {
```
lorisleiva, line 18: Why is this commented?

lorisleiva (changes requested):

packages/js/test/plugins/auctionHouseModule/directBuy.test.ts
lorisleiva: @zaxozhu Don't forget about this. The buyer should not be passed in if we want minimum input.

packages/js/src/plugins/auctionHouseModule/operations/createListing.ts
zaxozhu: I remember I had an issue when I was using an external seller. Will check again and add a comment.

packages/js/test/plugins/auctionHouseModule/directBuy.test.ts
zaxozhu: Thanks, I missed that

packages/js/src/plugins/auctionHouseModule/operations/directSell.ts
lorisleiva: Does it need it? If so can it be pushed to the operation handler instead of the transaction builder?

zaxozhu: Please, check again. I did a revisit to allow Listing to store assets without a token.

zaxozhu: That way it's only required to load up Purchase after the direct buy/sell

zaxozhu: It can be optimized not to load purchase when we have token provided, but I haven't done it yet.

packages/js/src/plugins/auctionHouseModule/operations/createListing.ts
```diff
@@ -407,7 +407,6 @@ export const createListingBuilder = (
     key.pubkey.equals(signer.publicKey)
   );
   sellInstruction.keys[signerKeyIndex].isSigner = true;
-  sellInstruction.keys[signerKeyIndex].isWritable = true;
```
lorisleiva: Can you leave this line and mark it as `false`? This attribute should not be undefined.

packages/js/src/plugins/auctionHouseModule/models/Listing.ts
```diff
@@ -30,8 +30,8 @@ export type Listing = Readonly<{
   /** A model of the Auction House related to this listing. */
   auctionHouse: AuctionHouse;
 
-  /** The Nft or Sft with the associated token account. */
-  asset: NftWithToken | SftWithToken;
+  /** The Nft or Sft related to this listing. */
+  asset: Sft | Nft | NftWithToken | SftWithToken;
```
lorisleiva, line 34: Why? Listings have to have a token account as there's no such thing as a Public Listing. We're losing valuable information here.

lorisleiva (changes requested):

packages/js/src/plugins/auctionHouseModule/operations/createListing.ts
zaxozhu: It's not undefined, the line above is to modify instruction keys to align it with provided signers

packages/js/src/plugins/auctionHouseModule/models/Listing.ts
zaxozhu: I wanted to fix the problem with missing token in direct sell due to public bid and skip the token loading

packages/js/src/plugins/auctionHouseModule/operations/createListing.ts
lorisleiva: Oh got it! Sorry about that.

lorisleiva (approved):
That's great, thanks!

PR #290, opened 2022-09-15
Fix typo in `UninitializedWalletAdapterError`

steveluscher:
Another option: renaming this to `wallet.connect()` since that's how folks have most likely called this API (`walletAdapterIdentity(wallet)`).


lorisleiva:
Thanks Steven!

PR #291, opened 2022-09-16
Auction House Partial Sale

zaxozhu:


packages/js/src/plugins/auctionHouseModule/operations/executeSale.ts
```diff
@@ -308,11 +320,15 @@ export const executeSaleBuilder = (
...
+  if (isPartialSale) {
+    const buyerPricePerToken = buyerPrice.basisPoints.div(
+      buyerTokensSize.basisPoints
+    );
 
-  // Data.
-  const price = isNative
-    ? lamports(bid.price.basisPoints)
-    : amount(bid.price.basisPoints, treasuryMint.currency);
+    if (!price.basisPoints.eq(buyerPricePerToken.mul(tokens.basisPoints))) {
```
zaxozhu, line 328: @lorisleiva This check copies the validation for the program side but adds more context to the user.

I don't have an idea how this should work from the Auctioneer side, since the listing price is the **u64::MAX** which requires the user to give **(u64::MAX / token size)** SOLs for each token. The same validation is in **auctioneer_execute_logic** and for me, it doesn't look right.

antey13 (approved):

packages/js/src/plugins/auctionHouseModule/operations/executeSale.ts
zaxozhu: Noticed in docs that partial sale isn't supported in Auctioneer

zaxozhu: https://docs.metaplex.com/programs/auction-house/auctioneer

packages/js/src/plugins/auctionHouseModule/errors.ts
```diff
@@ -134,10 +149,28 @@ export class WithdrawFromBuyerAccountRequiresSignerError extends AuctionHouseErr
     super({
       options,
       key: 'withdraw_from_buyer_account_requires_signer',
-      title: 'Withdraw From Buyer Account Requires Signer',
+      title: 'Withdraw From Buyer Account Requires Signer.',
```
lorisleiva, line 152:
```suggestion
      title: 'Withdraw From Buyer Account Requires Signer',
```

packages/js/src/plugins/auctionHouseModule/errors.ts
```diff
@@ -134,10 +149,28 @@ export class WithdrawFromBuyerAccountRequiresSignerError extends AuctionHouseErr
...
+/** @group Errors */
+export class PartialPriceMismatchError extends AuctionHouseError {
+  constructor(options?: MetaplexErrorOptions) {
+    super({
+      options,
+      key: 'partial_price_mismatch_signer',
+      title:
+        'The calculated partial price does not equal the partial price provided.',
+      problem:
+        'You are trying to execute the sale by providing the wrong price for the token.',
```
lorisleiva, line 169: Could we pass in the expected price and the actual price as error arguments so we can show them in the `problem` text? You can use `formatAmount` to display them as text.

packages/js/src/plugins/auctionHouseModule/operations/executeSale.ts
```diff
@@ -406,7 +436,7 @@ export const executeSaleBuilder = (
   // Receipt.
   const shouldPrintReceipt =
     (params.printReceipt ?? true) &&
-    Boolean(listing.receiptAddress && bid.receiptAddress);
+    Boolean(listing.receiptAddress && bid.receiptAddress && !isPartialSale);
```
lorisleiva, line 439: Interesting, does printing a purchase receipt not work with partial sales?

packages/js/src/plugins/auctionHouseModule/operations/executeSale.ts
```diff
@@ -308,11 +320,18 @@ export const executeSaleBuilder = (
...
+  }
+  if (isPartialSale) {
+    const buyerPricePerToken = buyerPrice.basisPoints.div(
+      buyerTokensSize.basisPoints
+    );
 
-  // Data.
-  const price = isNative
-    ? lamports(bid.price.basisPoints)
-    : amount(bid.price.basisPoints, treasuryMint.currency);
```
lorisleiva, line 315: Since we don't seem to be using prices as amounts anymore, is there a need to require the `treasuryMint` property of the `auctionHouse` parameter anymore? It might be used somewhere else but I just want to make sure we're not requiring more properties than we need to keep the "Minimum Valuable Input" principle.

packages/js/test/plugins/auctionHouseModule/helpers.ts
```diff
@@ -18,7 +18,5 @@ export const createAuctionHouse = async (
 
   await mx.rpc().airdrop(auctionHouse.feeAccountAddress, sol(100));
 
-  return {
-    auctionHouse,
-  };
+  return auctionHouse;
```
lorisleiva, line 21: Thanks for taking to time to refactor the tests. ❤️

packages/js/test/plugins/auctionHouseModule/executePartialSale.test.ts
```diff
@@ -0,0 +1,261 @@
...
+    })
+    .run();
+
+  // Then the user must receive 3 Tokens.
+  const buyerToken = await mx
+    .nfts()
+    .findByToken({ token: purchase.asset.token.address })
+    .run();
+
+  t.equal(buyerToken.token.amount.basisPoints.toNumber(), 3);
```
lorisleiva, line 69: Could we add more checks here?
- Checks on the state of the listing after the partial buy.
- Checks on the tokens owned by the seller.

packages/js/src/plugins/auctionHouseModule/operations/executeSale.ts
zaxozhu: Yes, the program expects an ExecuteSale instruction and doesn't support ExecutePartialSale. 

https://github.com/metaplex-foundation/metaplex-program-library/blob/master/auction-house/program/src/receipt/mod.rs#L472

As you can see it tries to deserialize the ExecuteSale instruction.

lorisleiva: Thanks for the update! 🍺

packages/js/src/plugins/auctionHouseModule/operations/executeSale.ts
zaxozhu: Yes, it's still needed to provide treasuryMint address to the instruction and also we need a full Auction House model to build a LazyPurchase model instance.

lorisleiva: Make sense, thanks for clarifying!

packages/js/test/plugins/auctionHouseModule/executePartialSale.test.ts
zaxozhu: Not sure what to test on the listing's state, I believe nothing changes there. 

Added a check for the seller's balance.

zaxozhu:
@lorisleiva Updated

packages/js/test/plugins/auctionHouseModule/executePartialSale.test.ts
lorisleiva: Hmm, how does the Program know that there are, say, two tokens left to sell on that listing? Because the seller could have hundreds of tokens for all we know but they only want to sell 5.

Do they use the `delegatedAmount` property of the token account? If so could we add a check on that?

packages/js/src/plugins/auctionHouseModule/errors.ts
```diff
@@ -159,14 +160,20 @@ export class WithdrawFromBuyerAccountRequiresSignerError extends AuctionHouseErr
 
 /** @group Errors */
 export class PartialPriceMismatchError extends AuctionHouseError {
-  constructor(options?: MetaplexErrorOptions) {
+  constructor(
+    options: Partial<MetaplexErrorOptions> & {
+      expected: Amount;
+      actual: Amount;
+    }
+  ) {
```
lorisleiva, line 168: Sorry to be a pain but that's not how the other errors add arguments. Can we do it this way so it's more consistent?

```suggestion
  constructor(
    expected: Amount,
    actual: Amount,
    options?: MetaplexErrorOptions
  ) {
```

packages/js/src/plugins/auctionHouseModule/operations/executeSale.ts
```diff
@@ -324,12 +324,16 @@ export const executeSaleBuilder = (
...
     const buyerPricePerToken = buyerPrice.basisPoints.div(
       buyerTokensSize.basisPoints
     );
 
-    if (!price.basisPoints.eq(buyerPricePerToken.mul(tokens.basisPoints))) {
-      throw new PartialPriceMismatchError();
+    if (!listingPricePerToken.eq(buyerPricePerToken)) {
+      throw new PartialPriceMismatchError({
+        expected: lamports(listingPricePerToken),
+        actual: lamports(buyerPricePerToken),
```
lorisleiva, line 335: This will not be lamports if the Auction House does not use SOL as a currency. I believe we have that information on the `auctionHouse` model.

packages/js/test/plugins/auctionHouseModule/executePartialSale.test.ts
```diff
@@ -61,12 +61,20 @@ test('[auctionHouseModule] execute partial sale on an Auction House', async (t:
...
-  t.equal(buyerToken.token.amount.basisPoints.toNumber(), 3);
+  t.equal(buyerTokens.token.amount.basisPoints.toNumber(), 3);
+
+  // And then the seller must have 2 Tokens left.
+  const sellerTokens = await mx
+    .nfts()
+    .findByToken({ token: listing.asset.token.address })
+    .run();
+
+  t.equal(sellerTokens.token.amount.basisPoints.toNumber(), 2);
```
lorisleiva, line 77: Has mentioned on the other thread let's also check the `delegatedAmount` here. And let's give more token to the seller initially to make sure we only offer the amount listed.

lorisleiva (changes requested):

lorisleiva (approved):
Thanks!

PR #312, opened 2022-09-27
Return signed transaction from `toTransaction()`

KartikSoneji:
The `toTransaction()` method returns an unsigned transaction.
This is inconvinient for actions that generate a keypair, for example minting an nft since the keypair needs to be extracted from te builder to sign the transaction separately.

This PR adds a backwards compatible extension to the method that returns a signed transaction if a `recentBlockhash` is passed to it.


lorisleiva:
Hi there 👋 

Thanks for raising this. I do believe transaction builders should be able to export Signed Transactions directly but I don't think this is the right approach so I will take the liberty to close this for now.

That being said, I am interested in understanding your use-cases more because you shouldn't need this PR to achieve what you are describing.

Every single operation that generates a Keypair allows you to provide it yourself so you can reuse it later in other transaction builders.

For instance, the `nfts().create()` operation has got a `useNewMint` parameter which you can use to provide your own generated Keypair ([See API References](https://metaplex-foundation.github.io/js/types/js.CreateNftInput.html)).

Additionally, once you have a transaction builder, you may send it directly using the RPC module like so.

```ts
await metaplex.rpc().sendAndConfirmTransaction(transactionBuilder);
```

Finally, here's an example combining multiple transaction builders to send within one transaction, Notice how the mint address is generated outside of the first transaction builder so we can use it in the second transaction builder.

```ts
// Get the transaction builder for the original NFT.
// Note we need to explicitly provide the mint Keypair
// to link the two transaction builders together.
const originalNftKeypair = Keypair.generate();
const originalNftBuilder = await metaplex
  .nfts()
  .builders()
  .create({
    uri: "https://arweave.net/123",
    name: "My NFT",
    sellerFeeBasisPoints: 500,  // 5% royalties.
    maxSupply: toBigNumber(10), // 10 limited editions.
    useNewMint: originalNftKeypair,
  });

// Get the transaction builder for the printed NFT.
const printedNftBuilder = await metaplex
  .nfts()
  .builders()
  .printNewEdition({
    originalMint: originalNftKeypair.publicKey
  });

// Merge the two transaction builders together.
const transactionBuilder = originalNftBuilder.add(printedNftBuilder);

// Execute the merged transaction builder.
await metaplex.rpc().sendAndConfirmTransaction(transactionBuilder);
```

I hope this helps.

HereFrank:
> Additionally, once you have a transaction builder, you may send it directly using the RPC module like so.
> 
> ```ts
> await metaplex.rpc().sendAndConfirmTransaction(transactionBuilder);
> ```
Hi!, besides what think Kartik I think that there are problems still with these section of the comment too...
I saw the source code of `sendAndConfirmTransaction(transactionBuilder);`

https://github.com/metaplex-foundation/js/blob/a013c4db37698120b593b0bfcc9e56351f327184/packages/js/src/plugins/rpcModule/RpcClient.ts#L109

The problem that I saw with the implementation of that function is that don't use the new `.sendTransaction()` and `confirmTransaction` methods... of the solanaWeb3 package u.u 
A thing which I read is important as is mentioned [here](https://github.com/solana-labs/solana/issues/23949) and [here](https://github.com/solana-labs/solana/pull/25227) 

Any way to update that function to use the new `connection.ConfirmTransaction` at least?
As I have read add good support for the timeout in the transactions...

lorisleiva:
That's a very good point! I'll make sure to update the SDK with the latest web3.js changes ASAP. Thanks for raising this!

Issue #345, opened 2022-10-11
Error while creating NFT


shiningsea0415:
I am trying to create NFT using metaplex-foundation/js module.
I installed the lastest versions.
But I cannot create even sample example NFT.

Here is the code.
let { nft } = await metaplex
      .nfts()
      .create({
        uri: "/fallbackImage.jpg",
        name: "New NFT",
        sellerFeeBasisPoints: 500
      });

First, I tried on testnet
But I got this error.
`{
	"jsonrpc": "2.0",
	"error": {
		"code": -32002,
		"message": "Transaction simulation failed: Error processing Instruction 4: Failed to serialize or deserialize account data: Unknown",
		"data": {
			"accounts": null,
			"err": {
				"InstructionError": [
					4,
					{
						"BorshIoError": "Unknown"
					}
				]
			},
			"logs": [
				"Program 11111111111111111111111111111111 invoke [1]",
				"Program 11111111111111111111111111111111 success",
				"Program TokenkegQfeZyiNwAJbNbGKPFXCWuBvf9Ss623VQ5DA invoke [1]",
				"Program log: Instruction: InitializeMint",
				"Program TokenkegQfeZyiNwAJbNbGKPFXCWuBvf9Ss623VQ5DA consumed 2967 of 1200000 compute units",
				"Program TokenkegQfeZyiNwAJbNbGKPFXCWuBvf9Ss623VQ5DA success",
				"Program ATokenGPvbdGVxr1b2hvZbsiqW5xWH25efTNsLJA8knL invoke [1]",
				"Program log: Create",
				"Program TokenkegQfeZyiNwAJbNbGKPFXCWuBvf9Ss623VQ5DA invoke [2]",
				"Program log: Instruction: GetAccountDataSize",
				"Program TokenkegQfeZyiNwAJbNbGKPFXCWuBvf9Ss623VQ5DA consumed 1622 of 1191670 compute units",
				"Program return: TokenkegQfeZyiNwAJbNbGKPFXCWuBvf9Ss623VQ5DA pQAAAAAAAAA=",
				"Program TokenkegQfeZyiNwAJbNbGKPFXCWuBvf9Ss623VQ5DA success",
				"Program 11111111111111111111111111111111 invoke [2]",
				"Program 11111111111111111111111111111111 success",
				"Program log: Initialize the associated token account",
				"Program TokenkegQfeZyiNwAJbNbGKPFXCWuBvf9Ss623VQ5DA invoke [2]",
[...]

PR #349, opened 2022-10-14
Fix Bundlr upload transaction failure in browser

zaxozhu:


lorisleiva (changes requested):

lorisleiva:
@zaxozhu I don't think that this fixes #345 btw.

I think #345 is due to the fact that they are trying to create an NFT on testnet and most of our programs are not deployed there (or on an old version).

lorisleiva (approved):

Issue #354, opened 2022-10-19
ReferenceError: AbortController is not defined


duythien0912:
Error when I try create NFT with getting-started-expressjs examples with

    "@metaplex-foundation/js": "^0.17.2",

```js
    const { uri } = await metaplex.nfts().uploadMetadata({
      name: "First NFT",
    });

    const { nft } = await metaplex.nfts().create({
      uri: uri,
    });

    console.log("[createNFT]: data " + nft.mint.toBase58());
    res.json(nft);
```

It throw error:

```
ReferenceError: AbortController is not defined
```

duythien0912:
Error because I used node 14 and only support node > 15
Switching to node 16 fixed this bug
But we should add node version required on the lib

PR #355, opened 2022-10-19
Specify the required Node.js version >= 16

duythien0912:
Specify the required Node.js version >= 16 for #354

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,6 +17,9 @@
   "repository": {
     "url": "https://github.com/metaplex-foundation/js.git"
   },
+  "engines": {
+    "node": ">=16.0"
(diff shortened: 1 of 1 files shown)


duythien0912:
@lorisleiva

lorisleiva:
Hi there 👋 This is already defined in the `package.json` of the SDK. https://github.com/metaplex-foundation/js/blob/main/packages/js/package.json#L70-L72

I'm not sure why we'd need to do this on the `package.json` of the monorepo.

duythien0912:
Somehow it still allow node 14 run project will close for now. 

Thank

PR #360, opened 2022-10-21
Documentation Error 'The amount of tokens to (mint -> send)'

CoachChuckFF:
The amount in the context of sendTokens should be amount to 'send' not amount to 'mint'


lorisleiva:
Thanks!

PR #370, opened 2022-10-24
feat: efficient findNftsBy methods by reducing RPC calls

KartikSoneji:
Both `findNftsByCreator()` and `findNftsByUpdateAuthority()` first fetch the nft mints, then make a call to `findAllByMintList()`.

https://github.com/metaplex-foundation/js/blob/8dcccc0a40938170db12827974552e66b47b84b9/packages/js/src/plugins/nftModule/operations/findNftsByCreator.ts#L92-L98

https://github.com/metaplex-foundation/js/blob/8dcccc0a40938170db12827974552e66b47b84b9/packages/js/src/plugins/nftModule/operations/findNftsByUpdateAuthority.ts#L76-L82

This is unnecessary since the `getProgramAccounts()` call returns all the needed data.
This PR allows parsing that data instead of making additonal RPC calls.


lorisleiva (approved):
Good shout! Thank you for your contribution. 🍺

PR #404, opened 2022-11-11
Return Metadata with Token from findNftsByOwner

fedorerd:
Reason: findNftsByOwner operation starts with retrieving all Token Accounts of provided owner. Having TAs we could easily retrieve Token model from them. But the operation is slicing data to keep only mintAddress from TA.

Change: instead of slicing TA data, keep full TA data and retrieve Token model from it. Add MetadataWithToken model and return it instead of Metadata. Also operation Output type was (Metadata | Nft | Sft | null)[] but in fact it could be only Metadata[] (after changes - MetadataWithToken[])

Usecase: now findNftsByOwner operation is enough to construct some instructions, for example transfer instruction without need to additionally search for TA again. Delegate amount checks also accessible just from findNftsByOwner operation result. Etc.

### DIFF
--- a/packages/js/src/plugins/nftModule/models/Metadata.ts
+++ b/packages/js/src/plugins/nftModule/models/Metadata.ts
@@ -194,3 +194,20 @@ export const toMetadata = (
 
 /** @group Models */
 export type MetadataWithToken = Metadata & { token: Token };
+
+/** @group Model Helpers */
+export const isMetadataWithToken = (value: any): value is MetadataWithToken =>
+  isMetadata(value) && 'token' in value;
+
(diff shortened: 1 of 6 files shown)


fedorerd:
`[41e245f](https://github.com/metaplex-foundation/js/pull/404/commits/41e245fae83fb8ef13cf346c31de48083ef2577a)` for #444

lorisleiva:
Hi there, thanks for your contribution but we've got no plans to merge this at the moment. Also, the `FindNftsByCreatorOutput` is purposefully looser to allow for plugins to define their own operation handlers and return more data than the default ones (e.g. indexers).

PR #406, opened 2022-11-13
Add Retries to findByMint call in CreateNFT operation

SaladAuthority:
createNftBuilder.sendAndConfirm() takes some time to process before the new data is returned from findByMint (eventual consistency)

This issue is only apparent for some RPCs like Quicknode. The public RPC endpoints devnet and mainnet are not impacted.

Adding retries fixed the issue

### DIFF
--- a/packages/js/src/plugins/nftModule/operations/createNft.ts
+++ b/packages/js/src/plugins/nftModule/operations/createNft.ts
@@ -5,7 +5,7 @@ import {
 import { Keypair, PublicKey } from '@solana/web3.js';
 import { SendAndConfirmTransactionResponse } from '../../rpcModule';
 import { assertNftWithToken, NftWithToken } from '../models';
-import { Option, TransactionBuilder, TransactionBuilderOptions } from '@/utils';
+import { Option, TransactionBuilder, TransactionBuilderOptions, RetryBuilder } from '@/utils';
 import {
   BigNumber,
   CreatorInput,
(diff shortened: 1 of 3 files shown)


SaladAuthority:
I also opened a customer support ticket to Quicknode, to find out the expected consistency model for their APIs and to find out if this is expected behavior

KartikSoneji:
https://github.com/metaplex-foundation/js/blob/7f8c5ebe2ef61cf6c86b0c965591e8cd90e1dd0e/packages/js/src/plugins/nftModule/operations/createNft.ts#L292-L296

`sendAndConfirm()` should ensure that the transaction is confirmed before trying to fetch the nft.
Can you double check that first?
Maybe try confirming the signature again or fetching the transaction?

Also, looking at the createNft logic, it doesn't make much sense to fetch the nft since all the information is already available to create the nft/sft object. Avoiding that can help reduce the number of RPC calls.

lorisleiva:
Hi there 👋 

Thanks for opening a PR. However, I agree with @KartikSoneji that a retry logic would not actually fix the race condition error because the `sendAndConfirm` method which delegates to web3.js is telling us that the transaction is confirmed even when asking for the `finalized` state.

I think the problem is that there might be a delay between the time the transaction is confirmed on the network and the time the NFT is indexed by an RPC like Quicknode.

I'm leaning to change the JS SDK's API such that operations either **write** or **read** but not both. It's a little less convenient but more predictable. It won't fix the race condition issue if you're fetching the NFT immediately after creating it but will push the problem to userland instead though.

In the meantime, you can achieve the same behaviour by using transaction builders instead.

```ts
const transactionBuilder = await metaplex.nfts().builders().create({ ... });
const { mintAddress } = transactionBuilder.getContext();
await metaplex.rpc().sendAndConfirm(transactionBuilder);

// Then, optionally fetch the NFT afterwards.
const nft = await metaplex.nfts().findByMint({ mintAddress });
```

PR #465, opened 2023-01-29
Add owner field (optional) to metadata and mint

thedonmon:
- Helps when finding by hashlist or any find operation who the owner of the mints are


lorisleiva:
Hi there, I'm afraid the account owner has nothing to do with the mint owner. `account.owner` refers to the program that owns the account, i.e. the SPL token program.

PR #474, opened 2023-02-14
Add docs on using Metaplex with your local validator

mikemaccana:


lorisleiva:
Sorry, I don't think this is the best place to document this as this is more Solana-related than Metaplex-related and doesn't make use of the Amman configuration we have in place. Thank you for understanding.

Issue #475, opened 2023-02-18
unable to update pNFT with rules set / need demo code for updating pNFTs once rule set


the-mercenaries-ltd:
According to the docs, we should be able to update pNFTs with something like:

```
await metaplex.nfts().update({
    nftOrSft,
    name: "My new NFT name",
    ruleSet: ruleSet.publicKey,
});
```

I already have Compatibility Rule Set assigned (AdH2Utn6Fus15ZhtenW4hZBQnvtLgM1YCW2MfVp7pYS5).

So I try simply using 
```
await metaplex.nfts().update({
    nftOrSft,
    name: "My new NFT name",
});
```

But both of these code snippets give the error from the TokenMetadataProgram:
```
  logs: [
    'Program metaqbxxUerdq28cj1RbAWkYQm3ybzjb6a8bt518x1s invoke [1]',
    'Program log: Instruction: Token Metadata Update',
    'Program log: Missing authorization rules account',
    'Program metaqbxxUerdq28cj1RbAWkYQm3ybzjb6a8bt518x1s consumed 21619 of 200000 compute units',
    'Program metaqbxxUerdq28cj1RbAWkYQm3ybzjb6a8bt518x1s failed: custom program error: 0x8a'
  ]
  ```
  
  I feel there must be a little more nuance needed to update a pNFT but after inspecting the codebase, I am unable to work it out.
  
  This code might be going in the right direction as it gives a different error:
  ```
    await metaplex.nfts().update(
    {
      nftOrSft: nft,
      name: "My new NFT name",
      authorizationDetails: {
        rules: new PublicKey("AdH2Utn6Fus15ZhtenW4hZBQnvtLgM1YCW2MfVp7pYS5")
      }
    },
  );
  ```

The error for this last snippet is 
```
  logs: [
    'Program metaqbxxUerdq28cj1RbAWkYQm3ybzjb6a8bt518x1s invoke [1]',
    'Program log: Instruction: Token Metadata Update',
    'Program log: Missing authorization rules program',
    'Program metaqbxxUerdq28cj1RbAWkYQm3ybzjb6a8bt518x1s consumed 16299 of 200000 compute units',
    'Program metaqbxxUerdq28cj1RbAWkYQm3ybzjb6a8bt518x1s failed: custom program error: 0x96'
  ]
  ```
  
  Does this make any sense to anyone?

htszn25:
[...]

PR #479, opened 2023-02-22
pNFTs: pass the authorizationRulesProgram in the instruction

the-mercenaries-ltd:
FIxes the broken update function in #475 for me.

Not tested the other functions but assuming the same fix carries to them.


lorisleiva (approved):
Thanks!

PR #480, opened 2023-02-22
[WIP] Add support for compressed NFTs

jnwng:
there's a lot of gaps to cover, so i figured it'd be worth getting a spot-check on the structure of this support before going too much further. this PR includes:
- the base-level requirements for `metaplex.nfts().create({ tree: <insert-tree-address>})` to mint a compressed NFT into an initialized Merkle tree and into a collection. 
- a thin wrapper around implementations of the Metaplex Read API for `metaplex.nfts().findByAssetId`

open questions that would be valuable to answer:
1. should the Read API wrapper be in this package? or perhaps in a different sub-package?

known things that are missing that i want to add, provided there's good feedback on the approach so far:
1. The ability to `transfer` NFTs
3. More Read API methods like `findByCreator`, `findByOwner`
4. Helper functions for initializing trees
5. NFTs without collections
6. Tests

---
this package was tested with a script that looks roughly like:
```TypeScript
  const connection = new ReadApiConnection(<insert-RPC-URL>);
  const metaplex = Metaplex.make(connection).use(keypairIdentity(wallet));

  const { tree } = await createTree(metaplex);

  const response = await metaplex.nfts().create({
    uri,
    name: 'The first!',
    sellerFeeBasisPoints: 500,
    tree: new PublicKey('Diz7dzSm9tknCUMhHDJnbBzQRpQtFoNjaxykKtzR9JFA'),
    collection: new PublicKey('3DJYB7uaGf5BjZLBasNu1WVrZwT8bZyptQKSshFDMAU3'),
    collectionAuthority: wallet,
  });
```
here's [a sample transaction](https://explorer.solana.com/tx/5wq98AaM8XqSFZnhnbP1M8Dt34hs8MnEwWPsXGsASvWhkiEtmXJpgYAWu1fia7K6tCrDriqeKNa2wipZWdS8LedW) that this code ran

### DIFF
--- a/packages/js/package.json
+++ b/packages/js/package.json
@@ -45,12 +45,14 @@
     "@bundlr-network/client": "^0.8.8",
     "@metaplex-foundation/beet": "0.7.1",
     "@metaplex-foundation/mpl-auction-house": "^2.3.0",
+    "@metaplex-foundation/mpl-bubblegum": "^0.6.2",
     "@metaplex-foundation/mpl-candy-guard": "^0.3.0",
     "@metaplex-foundation/mpl-candy-machine": "^5.0.0",
     "@metaplex-foundation/mpl-candy-machine-core": "^0.1.2",
     "@metaplex-foundation/mpl-token-metadata": "^2.7.0",
     "@noble/ed25519": "^1.7.1",
     "@noble/hashes": "^1.1.3",
+    "@solana/spl-account-compression": "^0.1.5",
     "@solana/spl-token": "^0.3.5",
     "@solana/web3.js": "^1.63.1",
     "bignumber.js": "^9.0.2",
@@ -63,7 +65,8 @@
     "lodash.isequal": "^4.5.0",
     "merkletreejs": "^0.2.32",
     "mime": "^3.0.0",
-    "node-fetch": "^2.6.7"
+    "node-fetch": "^2.6.7",
+    "typescript-collections": "^1.3.3"
   },
   "typedoc": {
     "entryPoint": "./src/index.ts",
--- a/packages/js/src/plugins/nftModule/operations/createCompressedNft.ts
+++ b/packages/js/src/plugins/nftModule/operations/createCompressedNft.ts
@@ -1,17 +1,21 @@
-import { TokenStandard, Uses } from '@metaplex-foundation/mpl-token-metadata';
+import {
+  TokenStandard,
+  Uses,
+  PROGRAM_ID as TOKEN_METADATA_PROGRAM_ID,
+} from '@metaplex-foundation/mpl-token-metadata';
 import {
   TokenProgramVersion,
   createMintToCollectionV1Instruction,
+  getLeafAssetId,
   PROGRAM_ID as BUBBLEGUM_PROGRAM_ID,
 } from '@metaplex-foundation/mpl-bubblegum';
 import {
   SPL_ACCOUNT_COMPRESSION_PROGRAM_ID,
-  // getConcurrentMerkleTreeAccountSize,
   SPL_NOOP_PROGRAM_ID,
-  // ConcurrentMerkleTreeAccount,
+  changeLogEventV1Beet,
 } from '@solana/spl-account-compression';
-import { PROGRAM_ID as TOKEN_METADATA_PROGRAM_ID } from '@metaplex-foundation/mpl-token-metadata';
 import { PublicKey } from '@solana/web3.js';
+import { BN } from 'bn.js';
 import { SendAndConfirmTransactionResponse } from '../../rpcModule';
 import { assertNft, Nft } from '../models';
 import { Option, TransactionBuilder, TransactionBuilderOptions } from '@/utils';
@@ -180,14 +184,6 @@ export type CreateCompressedNftInput = {
    * @defaultValue `false`
    */
   collectionAuthorityIsDelegated?: boolean;
-
-  /**
-   * Whether or not the provided `collection` is a sized collection
-   * and not a legacy collection.
-   *
-   * @defaultValue `true`
-   */
-  collectionIsSized?: boolean;
 };
 
 /**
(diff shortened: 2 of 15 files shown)


jnwng:
/cc @lorisleiva

i will continue to add support for some of the other functions while you're taking a peek at this, but let me know if you have any big changes you'd want to see here

packages/js/src/plugins/nftModule/plugin.ts
```diff
@@ -105,6 +107,7 @@ export const nftModule = (): MetaplexPlugin => ({
     op.register(createNftOperation, createNftOperationHandler);
     op.register(createSftOperation, createSftOperationHandler);
     op.register(deleteNftOperation, deleteNftOperationHandler);
+    op.register(findNftByAssetIdOperation, findNftByAssetIdOperationHandler);
```
lorisleiva, line 110: All new operations should be registered here in case someone decided to invoke them directly.

```suggestion
    op.register(findNftByAssetIdOperation, findNftByAssetIdOperationHandler);
    op.register(createCompressedNftOperation, createCompressedNftOperationHandler);
    op.register(transferCompressedNftOperation, transferCompressedNftOperationHandler);
```

packages/js/src/plugins/nftModule/operations/createCompressedNft.ts
lorisleiva, line 1: I see a lot of TODOs in this file. Is this operation ready enough to be merged as-is?

packages/js/package.json
```diff
@@ -63,7 +65,8 @@
     "lodash.isequal": "^4.5.0",
     "merkletreejs": "^0.2.32",
     "mime": "^3.0.0",
-    "node-fetch": "^2.6.7"
+    "node-fetch": "^2.6.7",
+    "typescript-collections": "^1.3.3"
```
lorisleiva, line 69: I don't see this dependency being used. Is that a peer-dependency requirement?

lorisleiva (changes requested):
Thanks both of you for your great work on this!

I've added a few questions/comments.

Additionally, all existing tests are passing except for the one that ensures the JS SDK can be imported as an ESM module. It seems that the `@solana/spl-account-compression` library is CJS only which can be imported to ESM modules but differently. See error below.

```
file:///home/runner/work/js/js/packages/js/dist/esm/plugins/nftModule/operations/createCompressedNft.mjs:3
import { SPL_ACCOUNT_COMPRESSION_PROGRAM_ID, SPL_NOOP_PROGRAM_ID } from '@solana/spl-account-compression';
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
SyntaxError: Named export 'SPL_ACCOUNT_COMPRESSION_PROGRAM_ID' not found. The requested module '@solana/spl-account-compression' is a CommonJS module, which may not support all module.exports as named exports.
CommonJS modules can always be imported via the default export, for example using:

import pkg from '@solana/spl-account-compression';
const { SPL_ACCOUNT_COMPRESSION_PROGRAM_ID, SPL_NOOP_PROGRAM_ID } = pkg;
```

The issue is, if we do change the way we import it as suggested, we break CJS modules. I've had this issue with the Bundlr library in the past and ended up having to create a helper function that removed the double `default` attribute from the package (See code below).

https://github.com/metaplex-foundation/js/blob/b3c9a450c968f26fbd970924527ccd4bc8f74ff2/packages/js/src/plugins/bundlrStorage/BundlrStorageDriver.ts#L37-L56

Would either of you mind taking a look at that?

packages/js/src/plugins/nftModule/plugin.ts
nickfrosty: Understood. Will update it to add each of the new operations

packages/js/package.json
nickfrosty: This can and will be removed. It is a peer dependancy for the spl compression sdk. Previously, it was not included in that package, but has since been added.

packages/js/src/utils/readApiConnection.ts
```diff
@@ -0,0 +1,268 @@
...
+  const currency: SplTokenCurrency = {
+    symbol: 'Token',
+    decimals: 0,
+    namespace: 'spl-token',
+  };
+
+  return {
+    model: 'mint',
+    address: new PublicKey(input.id),
+    // TODO(jon): Presumably, this should be the Master Edition address upon decompression
```
jnwng, line 60: @lorisleiva just want to confirm this. Read API does not provide this but we can derive it, if this is indeed supposed to be the Master Edition address that would show up upon decompression

packages/js/src/utils/readApiConnection.ts
```diff
@@ -0,0 +1,268 @@
...
+    mintAuthorityAddress: new PublicKey(input.id),
+    // TODO(jon): Presumably, this should be the Master Edition address upon decompression
```
jnwng, line 62: same with this one

packages/js/src/utils/readApiConnection.ts
```diff
@@ -0,0 +1,268 @@
...
+    tokenStandard: TokenStandard.NonFungible,
+
+    collection: collection
+      ? { address: new PublicKey(collection.group_value), verified: false }
+      : null,
+
+    compression: input.compression,
+
+    // TODO(jon): Read API doesn't return this info
+    collectionDetails: null,
```
jnwng, line 113: this is a restriction from the Read API standpoint, we should definitely get grouping info from the API

packages/js/src/utils/readApiConnection.ts
```diff
@@ -0,0 +1,268 @@
...
+    // TODO(jon): Read API doesn't return this info
+    uses: null,
```
jnwng, line 115: i'm not sure whether or not this is supported, but regardless its not coming through Read API. probably safe to null this out for now

packages/js/src/utils/readApiConnection.ts
```diff
@@ -0,0 +1,268 @@
...
+    // TODO(jon): Read API doesn't return this info
+    programmableConfig: null,
```
jnwng, line 117: not supported, so probably okay to null it out

packages/js/src/utils/readApiConnection.ts
lorisleiva: I think this is good enough for now. I'm a bit worried about setting this to the edition PDA since this would only make sense for non-fungible token standards.

lorisleiva (approved):

PR #507, opened 2023-05-05
Adding "authorizationDetails" to update()

fa8ster:


lorisleiva:
Thanks for this but unfortunately that updated example does not make a lot of sense within its context. It would make more sense to explain that in the PNFT section of the documentation.

PR #511, opened 2023-05-08
fix: fixed minting compressed NFTs

nickfrosty:
When attempting to mint a compressed NFT:

1. the "builder" method (e.g. `metaplex.nfts().builders().create()`) actually mints the compressed NFT, vice returning the builder object. This is not the desired action.

2. when attempting to create a compressed NFT using `metaplex.nfts().create()`, an error occurs due to changes in the spl-account-compression sdk. Specifically with deriving the `assetId` of the freshly minted compressed NFT.

This PR fixes these two issues and results in the correct `create` response for both


lorisleiva (approved):
Thanks!

